PR #1, opened 2017-06-16
Add editorconfig rule for markdown files

crashuniverse:
We like markdown files and this is a rule that makes our `.editorconfig` file respect markdown syntax of using 2 spaces for new lines.


james-nash:
Hah - I never knew about the 2 space thing in MD. Everyday's a school day! :-)

Merging - thanks for the contribution! Our very first contribution!! 🎉

PR #2, opened 2017-06-19
added stylelint

ThePeach:
Added Stylelint as dev dependency and added a basic configuration file for it.

There are some more in depth rules we can use, but at the moment this might suffice? Suggestions are welcome, maybe we can come up with a more strict set of rules before this gets merged.


james-nash:
Looks good.
Once we start adding build scripts, we need to remember to include some linting tasks that run this :-)

PR #3, opened 2017-07-06
Added Pattern Lab and SASS skeleton and build scripts

james-nash:
Copied the gulpfile from Matteo's PL starter project and slightly modified some of the file paths based on the comments in https://digitalrig.atlassian.net/wiki/display/GRAV/UI+lib+considerations


.gitignore
```diff
@@ -1,3 +1,9 @@
+# Exlclude build Output
+dist/
+
+
+######### App / OS-specific guff ##########
+
 ### macOS ###
 *.DS_Store
```
james-nash: Good spot.

This was auto-generated by https://www.gitignore.io/ . Weirdly, they don't have the `*` on other macOS ignores, so it looks deliberate. However, I can't find any information suggesting that the file is ever called anything other than just `.DS_Store`.

gulpfile.js
```diff
@@ -0,0 +1,306 @@
...
+ * The gulp wrapper around patternlab-node core, providing tasks to interact with the core library and move supporting frontend assets.
+******************************************************/
+const gulp = require('gulp');
+const path = require('path');
+const browserSync = require('browser-sync').create();
+const sass = require('gulp-sass');
+const eyeglass = require('eyeglass');
+const argv = require('minimist')(process.argv.slice(2));
+const chalk = require('chalk');
+const concat = require('gulp-concat');
```
james-nash: Are we not?
Sure, the location and name of `pattern-scaffolding.css` may change, but I'd have thought we will have one.

gulpfile.js
```diff
@@ -0,0 +1,306 @@
+/******************************************************
+ * PATTERN LAB NODE
```
james-nash: Yup. On it.

gulpfile.js
james-nash: Ah. Didn't quite understand this at first, but now I do. On it.

ThePeach:
👍

PR #4, opened 2017-07-12
Improved stylelint rules

ThePeach:
### Added
- CHANGELOG

### Changed
- fixed initial semver
- updated stylelint rules and dependencies
- updated README.md with information about versioning and release process


.stylelintrc
```diff
@@ -1,9 +1,75 @@
...
+    "scss/dollar-variable-no-missing-interpolation": true,
+    "scss/at-mixin-argumentless-call-parentheses": "never",
+    "plugin/selector-bem-pattern": {
+      "preset": "bem",
+      "presetOptions": {
+        "namespace": "ln"
+      },
+      "componentName": "(([a-z0-9]+(?!-$)-?)+)",
+      "componentSelectors": {
+        "initial": "\\.{componentName}(((__|--)(([a-z0-9\\[\\]'=]+(?!-$)-?)+))+)?$"
```
ThePeach: still unsure the regex and what effectively the `initial` value does (e.g. if I should add the `combined` one too), the docs are not particularly clear about it: https://www.npmjs.com/package/stylelint-selector-bem-pattern

james-nash: Tried reading the docs and unsure about this too. Does it work if we remove the `componentSelectors` option altogether. I think the default is BEM, and suspect that option's only needed if you wanted a custom, non-BEM-ish pattern for your class names.

.stylelintrc
```diff
@@ -1,9 +1,75 @@
...
+      }
+    ],
+    "scss/selector-no-redundant-nesting-selector": true,
+    "scss/at-extend-no-missing-placeholder": true,
+    "scss/dollar-variable-no-missing-interpolation": true,
+    "scss/at-mixin-argumentless-call-parentheses": "never",
+    "plugin/selector-bem-pattern": {
+      "preset": "bem",
+      "presetOptions": {
+        "namespace": "ln"
```
james-nash: Shouldn't our class namespace be something more Gravity-related? Perhaps `grav` or even just `gr` if you want to save keystrokes.

With the config as it is, it will expect class names to always begin `ln-*`

CHANGELOG
```diff
@@ -0,0 +1,14 @@
+# Changelog
```
james-nash: This file should be called `CHANGELOG.md`. That'll ensure that things like GitHub reliably render the markdown formatting. Also, it's what the FAQ at keepachangelog.com recommends :-)

james-nash:
@ThePeach I've added my review comments (apologies for the loooong delay!). Just a few minor things. Great work otherwise.

Thanks for documenting the release process! It reminded me: In due course we could look at automating some of that stuff - in the past I've used `gulp-bump` to simplify updating the version info in `package.json` and other files.

.stylelintrc
ThePeach: so, should I leave this out in the meantime, otherwise I would go for something unique like `gy` or `ms2` (as its SI unit of measure)?

PR #5, opened 2018-01-09
Gravy integration

james-nash:
Added Gravy as a dependency.


package.json
```diff
@@ -55,5 +56,12 @@
     "stylelint-order": "^0.5.0",
     "stylelint-scss": "^1.5.1",
     "stylelint-selector-bem-pattern": "^1.1.0"
+  },
+  "dependencies": {
+    "@buildit/gravy": "github:buildit/gravy#v2.0.0-alpha.2"
+  },
+  "engines": {
+    "node": "^8.9.4",
+    "npm": "^5.6.0"
```
ThePeach: is this really needed?
also, wouldn't it be better to mark node version as "~8.9.4" perhaps?

james-nash: `^` and `~` should be equivalent in this instance (see https://docs.npmjs.com/misc/semver).

PR #7, opened 2018-01-16
added nvm configuration

ThePeach:
what the title says


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
james-nash: Why not `lts/carbon`? Then it matches what's specified in `package.json`'s `engines` section.

If we eventually move the minimum required Node version to something newer, we can update both together

PR #8, opened 2018-01-17
added basic fonts configuration

ThePeach:
- upgraded gravy to alpha 3
- added base configuration of fonts for gravy
- updated json of fonts atom
- added import for Cabin and Roboto
- updated base HTML definition of fonts


src/ui-lib/sass/00-settings/typography.scss
```diff
@@ -1,7 +1,10 @@
 /**
   Typography-related values
 */
+@import url('https://fonts.googleapis.com/css?family=Cabin:400:600|Roboto');
 
 $font-size-base: unset;
-$line-height-default: 1.2;
+// TODO remove and use gravy `$base-font-size` and `$base-line-height`
```
james-nash: The existing variables were just placeholders. Please just go ahead and remove them and reference the Gravy-generated ones directly where needed.

src/ui-lib/sass/00-settings/gravy.scss
```diff
@@ -1,2 +1,22 @@
...
+      stack: ('Cabin', sans-serif),
+      weights: (
+        regular: 400,
+        bold: 600
+      )
+    )
+  )
+);
+
+@include gravy-init;
```
james-nash: Just an idea, but for developer convenience, it might be good to add a comment here with the URL of any Gravy docs that cover the set of vars/maps/mixins it generates at this point.

ThePeach:
Updated to Gravy Beta 1.
Killed the TODOs as suggested, moved the initialisation of gravy in the `heydon.scss` file, and added the "sensible typography defaults" that gravy provides. It makes me think if we should actually move some code output somewhere else.

Haven't added any Gravy documentation link, which might be better placed somewhere else perhaps?

james-nash:
I think calling those mixins in `_heydon.scss` is fine for now. If for some reason it later turns out not to be the best home, we can always move them.

PR #11, opened 2018-01-19
Gulp enhancements

james-nash:
* Split monolithic `gulpfile.js` into modules.
* Since we're using Gulp 4.0, I've made the modules export task functions rather than immediately registering them as tasks. This feels more flexible, so the gulpfile can then only register whatever tasks it actually needs to expose via the CLI.
    * A side effect of this is that several Pattern Lab task functions exist which are not being used. I decided to leave them in, in case we need them later. However, open to suggestions on this.
* I split the BrowserSync-related stuff out into a module of its own. My reasoning is that this is not really functionality provided by PL itself. Also, in PL3, it has it's own server/watcher built-in but it seems less sophisticated compared to BrowserSync. This way, when we upgrade to PL3, it'll be easier for us to continue using BrowserSync if we wish.
* I've put some of the paths to source and build dirs into `index.js`. In addition to our own build scripts, that info is useful for consumers of `gravity-ui-sass` such as the website. They'll be able to programatically get the correct path to our CSS to then copy to their own build output dir.


james-nash:
Cheers!

PR #12, opened 2018-01-22
Improved UI for colour palettes and swatches

ThePeach:
- reworked UI for swatches
- improved color data structure
- added WCAG AA data
- added relevant styles
- split colors into different patterns for more reusability
- updated stylelint class patterns

A bit of background, this is the porting of most of the work of UI that I did in the past for the starterkit with some minor improvements.
The way the colours are laid out can be improved and currently on mobile it just goes on a single column without much more than that. Suggestions are more than welcome.


src/styleguide/_patterns/00-particles/00-palettes/02-yellows.json
```diff
@@ -0,0 +1,89 @@
...
+          "dark": {
+            "small": "pass",
+            "big": "pass"
+          }
+        }
+      }
+    },
+    {
+      "color": {
+        "name": "Murican Cheese",
```
james-nash: Can we put the apostrophe back in?

**`Murican**

I think that makes the fact that it's derived from "**A**merican" clearer.

src/styleguide/css/pattern-scaffolding.css
```diff
@@ -1,65 +1,185 @@
...
+/** @define sg-swatch */
+.sg-swatch {
+  display: block;
+  height: 6em;
+  margin-bottom: 0.3em;
+}
+
+/** @define sg-swatch-check */
+.sg-swatch-check {
+  display: grid;
```
james-nash: CSS Grid. Yay!

_But_, is it not overkill in this instance, since it's a 1-dimensional layout? Flexbox could achieve the same thing.

Happy to accept this as it is, but if you fancy it/have spare time it might be worth considering simplifying this to be flex-based instead. Ditto for the nested `.sg-swatch-check__group` list.

ThePeach: I'm actually thinking of iterating a bit more on this so we can have a proper look on large display so it shows all the colours on the same row. Flexbox is a bit of a faff when it comes to that, while on the `.sg-swatch-check__group` is enough.
Let me see if I can get something done soon enough.

ThePeach:
> On a related note, given that the pattern-scaffolding.css is beginning to grow now, perhaps we should consider turning it into SASS and splitting it into partials. That would also allow us to potentially reference the same vars and mixins that gravity.css uses.
> 
> However, that's well beyond the scope of this PR, so I'll raise an issue for it and we can decide about that another time.

Indeed, agreed on this, please raise an issue, it would be worth doing that.

PR #15, opened 2018-01-23
WEB-64

brodiebuildit:
Updates to travis build for NPM publishing.
This is to republish what was in Master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,22 @@
 language: node_js
 node_js:
-  - "lts/*"
+  - lts/*
 cache:
   directories:
-    - "node_modules"
-
+    - node_modules
 env:
-  - PROD_TAG_REGEXP="^v[0-9]+\.[0-9]+\.[0-9]+.*" #Regexp for release (deploy to Prod) tags
-
+  global:
+    - NPM_TAG_REGEXP="^v([0-9]{1,3}\.){2}[0-9]{1,3}.*$"
+    - NPM_USER_EMAIL_ADDRESS="buildit.dev@gmail.com"
+    - NPM_USERNAME="buildit.dev"
+    - PROD_TAG_REGEXP="^v[0-9]+\.[0-9]+\.[0-9]+.*"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,34 +1,48 @@
...
+    - node_modules
 env:
-  - PROD_TAG_REGEXP="^v[0-9]+\.[0-9]+\.[0-9]+.*" #Regexp for release (deploy to Prod) tags
-
+  global:
+    - NPM_TAG_REGEXP="^v([0-9]{1,3}\.){2}[0-9]{1,3}.*$"
+    - NPM_USER_EMAIL_ADDRESS="buildit.dev@gmail.com"
+    - NPM_USERNAME="buildit.dev"
+    - PROD_TAG_REGEXP="^v[0-9]+\.[0-9]+\.[0-9]+.*"
+    - secure: "a9edwLnErePBD3w7nEEkEQBm4XGrFH0SCuQfHx9PHsANP2YTwI7CThmZsyqFYj0SH+h3y16l6jm1gU9HeFFJfbWS5x/JbCJi3U4sX0O6mxANmM5VTR+yC0J85Xzdn/GIHWy842ly1OtEVggtQt+TJPGZaZmjIwZxTrbQNlHprbSENvrJkggUa/x41iwst1C0UalTyhLt/bo2Co3xN7Ct+QOdz6CeBLSVvB0FT3yyONzRR9C8c56qHEh+rZtU1haUZKkEn+4esn72atQ8uVSmvOq9VU02HHsU04Sc2NdceLzrWkPTgpMxoX3ZpHb0EO5Kf6FMmqoPkNynmF/sXCm+zfblSTY/zgXm7BdrThNIFaLMGOSCY/apIibZcO4CimFjT9ohMR5tDs3tWRyrILmmHlU6A5w9BANIGwHiJYr3wn9l4vNBafXAnUTLQU0qWDYB3lcIAXHNiV6KZ4f3CNc/FRU9tf+yN2wxCX3JY5oPFLKZ7SJ6E+t7LpVUwdyxsP+qF9JkoiVp7pi4kUjEtfnT9cdg6PHQ2buRV6Oh1wPNup8+HgDL3Go5hfok8eDisX4V3mTYTlUt9LfJ/vIOIlNDkQ9ZcZm8KhtdGMjGSUKvd5GdLiI/LW2kdlNOXUpG5XvkN8BhqjJfQbihjU5g2D0vFtw4HArz+ZWAtV8xCZ1Dbxg=" #ENCRYPTD_NPM_AUTH_TOKEN environment variable
```
james-nash: Is it Ok to have tokens like this exposed in a publicly visible repo? Ditto for the other `secure` values below.

brodiebuildit: This is encrypted using the travis encrypt command which uses the public keys of the project to protect the data so that Travis can decrypt them using the private key.
This way just catalogues the values that are needed so that you don't have to list them in the Readme which runs the risk of being out of sync.
We can set up the values in Travis as an alternative, either way they are protected. 

Open to removing them from config and placing them in the server if this is preferred?

james-nash: Cool. It's absolutely fine as is then.

I was just curious. Every day is a school day! :-D

PR #20, opened 2018-02-02
Page header

james-nash:
Hey @ThePeach, here's the header work I've done so far. There's one outstanding thing that I wanted to add, but I've run out of time:

* Make the nav menu collapsible on narrow viewports
    * My plan was to write some simle JS to let the menu button toggle an `aria-expanded="true|false"` attribute on the `<nav>` element and then use CSS to apply an open/close animation or transition.
    * I was reading stuff like https://inclusive-components.design/menus-menu-buttons/ to find a solution that will play nice with assistive tech.

Other than that, it's all good to go though. Here are the main things I've done (which'll hopefully aid your review):

## SASS conventions
I had a stab at establishing some naming & coding conventions for the SASS code (see https://github.com/buildit/gravity-ui-sass/blob/page-header/docs/naming-conventions.md).

It's using HTML element & attribute selectors as much as possible and BEMIT for everything else.

A tad unusual perhaps, but I see a lot of people who are unfamiliar with HTML5's many built-in features and semantics and was hoping this might be a good way to raise awareness of those things.

However, if you feel strongly, you're welcome to adapt or modify these.

For now, I've tried to follow these conventions for my own code. And, FWIW, it's gone pretty smoothly so far.

## SVG icons
I'm using inline SVG symbols and then referencing them by ID wherever icons need to be displayed in the UI. It's all wired up into PL.

I suggest pulling my branch, running locally and then reading the pattern info descriptions I added. Hopefully they explain what I did and why.

## Page header
With the coding conventions and icon system in place, I've begun building all the parts I needed in order to assemble a header. I produced some extra atoms along the way (e.g. types of buttons).

I've made an effort to add solid documentation to each pattern to encourage correct usage and (hopefully) teach people a thing or two too. :-)

The `nav-menu` component uses Flexbox and I deliberately populated it with awkward content (long link labels & more links than fit in a single row) to stress test it. My intention was to make it as robust as possible.

The example page I added shows it populated with the links we'd _actually_ use for the site.

Aesthetically it's basic for now. I think we should build out a skeletal UI in Pattern Lab first and _then_ begin tweaking and tuning the visuals.


ThePeach:
This is just the first of my review of this pull request, as I'm still going through it and it would take me a bit to finish it.

## Code standards

Very good job 👍 , I think for now they work, as you said it's just an initial draft and I'm generally OK with it, both in using BEMIT and few additional stuff in the way selectors are used.
My main problem here is the amount of variables being generated, which IMO will soon become an impediment as we scale. The idea is that we should strive to keep them as easy to remember, and at the same time functional. A very good example is the amount of colours defined and aliased. Which is why probably having some mixins that give you the right colour might be a better choice (especially if you're looking for the accessible _inverted_ colour for any given one.... (e.g. I want an orange tint that can go on black). Food for thought.

Regarding the various patterns that you created in the way of creating the navigation organism and the header in general I think you went a bit overboard. Nothing bad with it, but there's a lot of stuff we are not yet using which is why I'm struggling a bit to understand if it actually makes sense, while I would have probably created the barely minimum rather than having to go back and refactor. Just a thought as well. I think it's fine so far.

Same goes for the styling: I would have probably avoided making the menu this much styled, but I think it's fine for now.I already see that font sizes and line heights need to be adjusted; I'm saying this mostly because now I'm drawn into the trick of seeing something visual and catching a few things that don't really work IMO. An example is the colours used on the `active` state of the links.

ThePeach:
I've noticed that some patterns are not being generated appropriately, I will try to double check this, but at the moment I had to go and re-save each pattern to have the output correctly generated.
This happened with the SVG icons, which didn't generate anything, and the menu links which outputted some `undefined` for some reason.

ThePeach:
I've cleaned up some patternlab styles and markup to reuse a bit of the existing classes. I think we might need to write some documentation or either start switching to SASS compilation altogether, before it starts to become a mess. As a precaution I've started adding some `@define` before each class definition.

ThePeach:
Regarding the collapsible menu, which is now the only thing I'd add, I wanted to implement it purely in HTML/CSS, [similarly to this CodePen](https://codepen.io/luxonglassing/pen/fugyq). TL;DR: the icon is basically a checkbox, that triggers the visibility of the menu itself.

PR #26, opened 2018-02-19
Hiring process

ThePeach:
- added styling for section
- ~~added `is-hidden` helper class~~


src/styleguide/_patterns/04-templates/01-careers.mustache
```diff
@@ -1,10 +1,22 @@
...
+  <p>
+    <a href="#the-hiring-process">The hiring process</a>
+  </p>
 
   {{#locations}}
     {{> organisms-job-list }}
   {{/locations}}
 
+  <section>
+    <a name="the-hiring-process"></a>
```
danbull: Could the `name` not be added to another element rather than having an empty anchor?

src/ui-lib/sass/06-utilities/_is-hidden.scss
```diff
@@ -0,0 +1,3 @@
+.is-hidden {
```
danbull: Is this being used? If so I think it needs to be name `.grav-u-is-hidden` and perhaps have `!important` since it will be a trump.

ThePeach: not actually used, I think I'm going to remove it for now 👍

ThePeach:
I'm also questioning the look and feel of this class: I was wondering if maybe just having the number and removing the part "Step" will make it more usable ❓

PR #27, opened 2018-02-21
Web 51

danbull:
- Added location card molecule and used in an organism, template, page
- Added temp city images and gulp copy task
- Created a WIP css grid layout, possibly it's too 'rigid'?
- Added `height: auto;` for images


gulp/ui-lib.js
```diff
@@ -70,9 +70,13 @@ function copyJsTask() {
 copyJsTask.displayName = taskNamePrefix + 'js';
 copyJsTask.description = 'Copies JS files.';
 
+function copyImgTask() {
+  return gulp.src(`${paths.srcImgDir}/**/*.{gif,jpg,png,svg}`)
```
ThePeach: be careful with `paths.srcImgDir`: even if the task per se works OK, these specific images will be also published on npm once we have merged. I think these images should be part of PatternLab directory (configuration file of PatternLab should tell you where the paths for the images should go IIRC)

gulp/ui-lib.js
```diff
@@ -70,9 +70,13 @@ function copyJsTask() {
 copyJsTask.displayName = taskNamePrefix + 'js';
 copyJsTask.description = 'Copies JS files.';
 
+function copyImgTask() {
```
ThePeach: the task should be renamed so that it's clear that it will copy images that will go into the UI library. (see following comment)

src/styleguide/_data/data.json
```diff
@@ -32,14 +32,31 @@
...
   },
+  "image": {
+    "path": "http://via.placeholder.com/425x600/000000/ffffff",
+    "alt": ""
+  },
   "jobs": [
     {}, {}, {}
+  ],
+  "offices": [
+    {}, {}, {}, {}, {}, {}
```
ThePeach: It would probably be better to have more, to see how it will adapt if we expand ❓

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
```diff
@@ -0,0 +1,17 @@
+<div class="grav-c-location-card">
+  <div class="grav-c-location-card__content">
+    <h2>{{office.name}}</h2>
+    <p>{{#office.addressline1}}{{office.addressline1}}<br>{{/office.addressline1}}
```
ThePeach: I'm wondering if you perhaps can avoid doing each single check for each property and only have it for those that we know could be missing (`addressline2` for instance), and wrap the whole thing in `{{#office}}{{/office}}` ? fair assumption?

src/styleguide/_data/data.json
```diff
@@ -32,14 +32,31 @@
     "city": "Visakhapatnam",
     "country": "India"
   },
-  "locations": [
-    {}, {}, {}
-  ],
+  "office": {
```
ThePeach: wouldn't be better to reuse the `locations` and `location` key that are used in the careers page?

gulp/ui-lib.js
danbull: I've removed this, it's now handled by patternlab

src/styleguide/_data/data.json
danbull: Added a couple more

src/styleguide/_data/data.json
danbull: Refactored that now

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
```diff
@@ -0,0 +1,17 @@
+<div class="grav-c-location-card">
+  <div class="grav-c-location-card__img">
+    <img src="{{location.img}}" width="425" height="300" alt="">
```
ThePeach: It doesn't seem like `width` and `height` attributes are needed at all.
You might also want  to fill in the `alt` attribute with some random copy with {{location.city}} ?

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
```diff
@@ -0,0 +1,17 @@
...
+  </div>
+  <div class="grav-c-location-card__content">
+    <h2>{{location.name}}</h2>
```
ThePeach: what's the difference between `{{name}}` and `{{city}}` ?

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
```diff
@@ -0,0 +1,17 @@
...
+    <p>{{#location.addressline1}}{{location.addressline1}}<br>{{/location.addressline1}}
```
ThePeach: do we need all these checks? Can we _safely_ assume they might not be there? Or can we assume that some are always there?

src/styleguide/_patterns/03-organisms/01-careers/00-job-list.mustache
```diff
@@ -1,11 +1,11 @@
 <section class="job-list">
-  <h2>{{location.city}}, {{location.country}}</h2>
+  <h2>{{location.name}}, {{location.country}}</h2>
```
ThePeach: see my previous comment.. you kept `city` and added `name`, so maybe it's the case of removing one of the two to avoid confusion?

src/styleguide/_patterns/05-pages/01-careers.json
```diff
@@ -19,6 +19,7 @@
       }
     },
     {
+      "url": "../../patterns/05-pages-02-locations/05-pages-02-locations.rendered.html",
```
ThePeach: I think you can just use `link.pages-locations` and that should be rendered for you

src/ui-lib/sass/04-objects/_locations-layout.scss
```diff
@@ -0,0 +1,19 @@
+.grav-o-locations-layout {
+  @media (min-width: gravy-breakpoint(regular)) {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
```
ThePeach: missing closed braket.
you want to have:
```
@media {
}
@media {
} 
```
since they specify `min-width`, the larger will get settings from the smaller.

src/ui-lib/sass/05-components/02-molecules/01-cards/_location-card.scss
```diff
@@ -0,0 +1,47 @@
+.grav-c-location-card {
+  position: relative;
+  margin: 0;
+  overflow: hidden;
+}
+
+.grav-c-location-card__content {
```
ThePeach: I've made full use of nesting in situations like these, so I don't know if you split elements styles aside or if it's just a question because of the limit of nesting depth...

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
danbull: In this case `height` and `width` probably won't make any difference, added out of habit. Can add some alt text in the data

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
danbull: The main reason for this is New York/Brooklyn. We call it New York but the address uses Brooklyn where there would usually be a city

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
danbull: Because we don't want the `<br>` to render if it's empty

src/ui-lib/sass/05-components/02-molecules/01-cards/_location-card.scss
danbull: You mean `&__content`? I prefer that but used to people hating it so didn't do it.

src/ui-lib/sass/04-objects/_locations-layout.scss
danbull: whoops

ThePeach:
go and merge this bitch up! 👯

PR #28, opened 2018-02-21
added namespaces to newly introduced classes

ThePeach:
Completely forgot about BEMIT. This PR is just to fix what has been already introduced, nothing more than a bunch of prefixes added to the classes. Also fixed ordering of attributes in classes.


danbull:
LGTM 👍  after conflicts are resolved

PR #35, opened 2018-03-06
added utility class for layout container

ThePeach:
- created `grav-l-container` mixin
- created `grav-l-container` class
- removed style from `main` element
- refactored markup to use new utility class
- added mixin to header and footer
- cleaned up header
- cleaned up footer

### DIFF
--- a/src/styleguide/_patterns/04-templates/00-page-layouts/00-basic-page.mustache
+++ b/src/styleguide/_patterns/04-templates/00-page-layouts/00-basic-page.mustache
@@ -1,5 +1,5 @@
 {{> organisms-page-header }}
-<main>
+<main class="grav-l-container">
   <h1>{{title}}</h1>
   {{#listitems.five}}
     <p>{{ excerpt.long }}</p>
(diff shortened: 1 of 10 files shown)


src/ui-lib/sass/06-utilities/_container.scss
```diff
@@ -0,0 +1,4 @@
+// just an utility class to use the grav-l-container mixin directly as a class.
+.grav-l-container {
+  @include grav-l-container;
+}
```
danbull: I think the naming is slightly off here. I'd suggest this is probably an `object` so should live there and be named `.grav-o-container`.

ThePeach: good point, I was unsure as well!

src/styleguide/_patterns/04-templates/00-page-layouts/00-basic-page.mustache
```diff
@@ -1,5 +1,5 @@
 {{> organisms-page-header }}
-<main>
+<main class="grav-l-container">
```
danbull: This class needs to be `grav-o-container`

PR #37, opened 2018-03-07
updated style of job-list

ThePeach:


danbull:
I really think the CTA is being overused here; it's losing it's role as a CTA and is just becoming the default style for a link on this page.  The office location could be a link maybe? Can discuss!

Also maybe instead of repeating "full time" this could be assumed unless it says otherwise? e.g. (Contract)

src/ui-lib/sass/05-components/03-organisms/_job-list.scss
```diff
@@ -1,12 +1,29 @@
...
+  }
+
+  > header > p {
     @include typi(small-text);
-    margin-top: $grav-sp-m;
+    margin-top: 0;
   }
 
   > ul {
     padding-left: $grav-sp-m;
```
danbull: I think this should be `0` on mobile too?

ThePeach: correct!!!!

ThePeach:
very good points, and to reply to that:

1) we don't really have anything to link to that, I'm going to remove that, and I do agree with you, it should be a normal link, not a CTA.
2) the full-time etc, maybe that's something we can check when doing the user test, as I think we would have the same problem if all the positions were contracts.

PR #38, opened 2018-03-09
Web 98

danbull:
- added ascii art atom
- added centred text modifier for sections
- creating error page template and page for 404


src/styleguide/_patterns/01-atoms/01-block-text/05-ascii-art.mustache
```diff
@@ -0,0 +1 @@
+<pre class="grav-c-ascii-art" aria-hidden="true">(╯°Д°)╯︵ ┻━┻</pre>
```
ThePeach: the ascii art should probably be extracted in a variable in the `_data.json` file

src/styleguide/_patterns/04-templates/00-page-layouts/01-error-page.mustache
```diff
@@ -0,0 +1,11 @@
+{{> organisms-page-header }}
+<main class="grav-o-container">
+  <section class="grav-o-section--centered">
+    <h1>{{> atoms-ascii-art }} {{title}}</h1>
```
ThePeach: I think we briefly talked about this, now that I see how it's been organised, I think it would be worth putting the ascii art in a separate container (a `<codeblock>` or a `<p>` with a special class?) and leave the error name in the `<h1>` so you don't have to do weird stuff.
You can always use the `@include typi('large-text')` if you need a larger font when styling

src/styleguide/_patterns/04-templates/00-page-layouts/01-error-page.mustache
```diff
@@ -0,0 +1,11 @@
...
+    <p>{{excerpt.short}}</p>
+    <div class="grav-o-container">
+      <a href="#the-hiring-process" class="grav-c-cta">Go to the homepage</a>
```
ThePeach: I think the `href` is wrong? can we put stuff in variables?

src/styleguide/_patterns/05-pages/00-page-layouts/01-error-page.json
```diff
@@ -0,0 +1,43 @@
+{
+  "title": "404",
```
ThePeach: I would use `404 - Page not found`

src/ui-lib/sass/05-components/01-atoms/01-block-text/_05-ascii-art.scss
```diff
@@ -0,0 +1,4 @@
+.grav-c-ascii-art {
+  font-size: 2rem;
```
ThePeach: see above, use `@include typi(large-text);`

src/styleguide/_patterns/04-templates/00-page-layouts/01-error-page.mustache
ThePeach: I forgot that your ascii art is already in a separate tag. I'd just say that probably it's better to split them up.

src/styleguide/_patterns/04-templates/00-page-layouts/01-error-page.mustache
danbull: Whoops! I'll fix that

src/styleguide/_patterns/05-pages/00-page-layouts/01-error-page.json
danbull: I changed that but then changed my mind as I thought most people visiting would be reasonable tech savvy

ThePeach: I wouldn't expect that, given a good percentage of visitors are supposedly clients which might not be.

src/styleguide/_data/data.json
```diff
@@ -5,6 +5,9 @@
     "medium" : "Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.",
     "long" : "Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
   },
+  "preformatted-text": {
+    "ascii-table-flip": "(╯°Д°)╯︵ ┻━┻"
```
ThePeach: The variable name is too specific, in case you want to override its value in another page, just use `preformatted-text` would be ok if you think that `performatted-text` would have more than one prop, then use a more generic one like `ascii` ?

PR #39, opened 2018-03-15
Fixes job list alignment

franz899:
- It fixes the first li inside a grav-c-job-list > ul

### DIFF
--- a/src/ui-lib/sass/05-components/03-organisms/_job-list.scss
+++ b/src/ui-lib/sass/05-components/03-organisms/_job-list.scss
@@ -25,7 +25,7 @@
     }
   }
 
-  li {
+  li:not(:first-child) {
     margin-top: $grav-sp-s;
   }
 }
--- a/src/ui-lib/sass/05-components/03-organisms/_job-list.scss
+++ b/src/ui-lib/sass/05-components/03-organisms/_job-list.scss
@@ -1,10 +1,12 @@
 .grav-c-job-list {
-  margin-top: $grav-sp-xl;
+  * + & {
+    margin-top: $grav-sp-xl;
+  }
 
   @media all and (min-width: gravy-breakpoint(regular)) {
     display: grid;
-    grid-template-columns: 30% 1fr;
     grid-gap: $grav-sp-m;
+    grid-template-columns: 30% 1fr;
   }
 
   > header > p {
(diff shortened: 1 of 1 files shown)


src/ui-lib/sass/05-components/03-organisms/_job-list.scss
```diff
@@ -25,7 +25,7 @@
     }
   }
 
-  li {
+  li:not(:first-child) {
```
danbull: I think this is fine but I've seen `li + li` (or similar) being used also. Is there a convention @ThePeach @james-nash?

ThePeach: 👍 second the `li + li`. Btw no convention here.

franz899: I think that `:not` follows more the css standards, if there is a rule that does exactly what we need why should we use another way, even if it accomplish the same thing.

I have no strong opinion about it.

ThePeach: They both follow standard, only difference is that `li + li` is a CSS pattern that follows the lobotomised owl pattern and is quite widely used.

franz899: Those are techniques though, not specific rules.

I know about the owl, and it is useful, but is a bit tricky to respect as a rule. There idea is to not to specify the margin on the single elements, but it has already been broken in this css.

If we want to keep it consistent then I propose this:
```
.grav-c-job-list {
  @media all and (min-width: gravy-breakpoint(regular)) {
    display: grid;
    grid-gap: $grav-sp-m;
    grid-template-columns: 30% 1fr;
  }

  > header > p {
    @include typi(small-text);
    margin-top: 0;
  }

  > ul {
    padding-left: 0;
    list-style-type: none;

    @media all and (min-width: gravy-breakpoint(small)) {
      columns: 2;
    }

    @media all and (min-width: gravy-breakpoint(regular)) {
      margin-top: 0;
    } 
  }

  li + li {
    margin-top: $grav-sp-s;
  }
}

* + .grav-c-job-list {
  margin-top: $grav-sp-xl;
}
```

ThePeach: sounds good to me, the last one can be written inside the main class as:

```
.grav-c-job-list {
  ...

  * + & {
    margin-top: $grav-sp-xl;
  }
}
```

franz899: Oh yeah of course, I copied my rule from Firefox.

PR #42, opened 2018-03-19
Web 116 - Create alternative version of locations page

franz899:
- New page style
- Locations static image

Let me know what do you think about rounder corners. I think the page looks better with them.

The images are created from Google maps' static API


ThePeach:
Quick one before I dig deeper into the code: the idea behind this ticket is to create an _alternative_, which is _another_ version of the location page, so we can submit that to the users to see which one they prefer. So I'd suggest to create a `pages/03-locations-alt.mustache` or something of the likes, together with an `location-card-alt`.

Aside from that, I was looking on how the the location cards were working, and I've got a couple of questions around them:

1. Is there any reason for using `grid` rather than, e.g. `float`ing the image? Given the image is of fixed size, the rest of the text shouldn't be a problem. As it is now, the whole configuration is a bit spread around and seems a bit of a stretch in the use of grid specifically in this instance.
2. If we use `grid` shouldn't it be better to use `auto` in the first column instead of splitting the content precisely at 50%? seems to be able to scale a bit better.
2a. maybe you should have had row templates as well as column templates for what you wanted to achieve.
3. from an accessibility/usability PoV, I think a link to the actual map would be good, maybe using the image, which due to the way it looks, feels like the natural candidate and as a textual link perhaps?

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
```diff
@@ -1,17 +1,11 @@
...
-    {{> atoms-link:grav-c-block-link }}
-  </div>
+  <h2>{{location.name}}</h2>
+  <address>
+    {{#location.addressline1}}{{location.addressline1}}<br>{{/location.addressline1}}
+    {{#location.addressline2}}{{location.addressline2}}<br>{{/location.addressline2}}
+    {{#location.city}}{{location.city}}<br>{{/location.city}}
+    {{#location.postcode}}{{location.postcode}}<br>{{/location.postcode}}
+    {{#location.country}}{{location.country}}{{/location.country}}
+  </address>
```
danbull: This is not the correct use of the address tag (you aren't the first and definitely won\t be the last) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
```diff
@@ -1,17 +1,11 @@
...
+  <img src="{{location.img}}" alt="{{location.img-alt}}">
```
danbull: The image should surely be inside an anchor and point to google maps? I'm a bit unsure about using a screen grab of google maps anyway, It's misleading and looks like it should behave like Google maps. Especially if you using touch gestures.

danbull: ...having said that, it's exactly what Mozilla do! e.g. https://www.mozilla.org/en-US/contact/spaces/mountain-view/

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
franz899: Totally forgot about that.

src/styleguide/_patterns/02-molecules/02-locations/location-card.mustache
franz899: Is not a screen grab from Google Maps. Is an API made to show a map but to avoid loading the JS bundle if not necessary. The images can be cached and they do the job, especially in a page like this where we want to show many different ones.

The anchor tag is ready now, I wanted to receive feedback on the design asap.

danbull: 👍

src/styleguide/_patterns/05-pages/03-locations-alt.json
```diff
@@ -0,0 +1,252 @@
...
+        "postcode": "Bangalore 560100",
+        "country": "India",
+        "img": "../../images/map/bengaluru.jpg",
+        "img-alt": "Bengaluru static map",
+        "map-link": "https://www.google.com/maps/place/Wipro+Limited/@12.838566,77.6592042,16z/data=!4m5!3m4!1s0x0:0x1d0caf77fe02554f!8m2!3d12.8384852!4d77.6571581?hl=en-US"
+      },
+      "control": {
+        "label": "Show on a map"
+      },
+      "jobs": [
```
ThePeach: all `jobs` entries can be removed as they won't affect the page and will make the JSON a bit more readable! 😄

src/ui-lib/sass/04-objects/_locations-layout-alt.scss
```diff
@@ -0,0 +1,9 @@
+.grav-o-locations-layout-alt {
+  display: grid;
```
ThePeach: this doesn't seem to be needed when it's on a single column.

src/ui-lib/sass/05-components/02-molecules/01-cards/_location-card-alt.scss
```diff
@@ -0,0 +1,39 @@
+.grav-c-location-card {
+  display: grid;
+  grid-gap: 1em;
```
ThePeach: I think `$grav-sp-m` is what you need here

src/ui-lib/sass/05-components/02-molecules/01-cards/_location-card-alt.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  @media (min-width: gravy-breakpoint(small)) {
+    grid-template-areas:
+      'title title'
+      'map info';
+    grid-template-columns: 1fr 1fr;
+  }
+
+  @media (min-width: gravy-breakpoint(medium)) {
+    margin-top: 2em;
```
ThePeach: I think `$grav-sp-l` is what you need here

src/ui-lib/sass/05-components/02-molecules/01-cards/_location-card-alt.scss
```diff
@@ -0,0 +1,39 @@
...
+    grid-area: info;
+  }
+
+  a {
+    margin: 0;
+    grid-area: map;
+  }
+
+  img {
+    border-radius: 5px;
```
ThePeach: please move the `5px` over to a variable.

james-nash:
Currently the heading of one card is closer to the image of the previous card than it's own. This creates the mistaken impression that the previous image is associated with the next heading.

E.g. in the screenshot below, the title "Dublin" looks like it belongs to the map above it, but that's actually for Denver:

image

The obvious fix would be to add a sufficiently large vertical margin between successive cards. Interestingly, the layout for larger screens (with 2 cards per row) already does this.

james-nash:
Also, some of the map markers aren't pointing to the right place (perhaps they're showing the middle of the postcode area instead of a specific building?).

For example, in the London map half of our 2FA building is visible on the right edge of the image, but map marker is 2 blocks West of there.

Similarly, in New York, it's pointing to the middle of a road but it should be the "WeWork Dumbo Heights" building to the East of there.

james-nash:
@franz899 I think the rounded corners look nice, _but_ it's a new element in our existing design style. So far everything is quite pointy and angular.

However, with @ThePeach's incoming new logo, we'll have something else that's rounded. :-)

So, I suggest we keep the rounded corners for now (to get more feedback), but perhaps consider what else in our UI could/should also be rounded.

franz899:
I've double checked the map coordinates and I use the same Lat Lon for static and live maps.

ThePeach:
Re the map display: I've checked and the link points you to the right place, but the static map is wrong:

static:
london

live map (opened through the link):
screen shot 2018-03-21 at 09 20 57

ThePeach:
@franz899 I wasn't really doubting that you didn't save the right images, I was just implying that that's apparently a bug on their end... I don't know whether this last commit the best way to deal with it on our end ❓ suggestions? @james-nash?

PR #44, opened 2018-03-19
WEB-131 - mirrored jobs from current website

ThePeach:
- cleaned up careers page json
- organised cities by country code


ThePeach:
good point, merging

PR #45, opened 2018-03-20
updated buildit logo

ThePeach:


franz899:
Shouldn't the small path align with the orange one? Now is slightly off.

ThePeach:
thank you for picking this one up, I've fixed it and noticed that I didn't replace the standalone icon, so that should be sorted now too.

PR #48, opened 2018-03-21
Rounded corners for buttons

franz899:
**This is not a card from JIRA**

I've created the PR to gather feedback around rounded corners for buttons and step lists. Let me know what do you think 😄 

### Things to have a look at:
- [http://localhost:3000/?p=viewall-atoms-buttons](http://localhost:3000/?p=viewall-atoms-buttons)
- [http://localhost:3000/?p=viewall-atoms-navigation](http://localhost:3000/?p=viewall-atoms-navigation)
- [http://localhost:3000/?p=viewall-molecules-lists](http://localhost:3000/?p=viewall-molecules-lists)

### DIFF
--- a/src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss
+++ b/src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss
@@ -10,6 +10,7 @@
     &:visited {
       background-color: $grav-co-interactive-secondary;
       border-color: $grav-co-interactive-secondary;
+      border-radius: $grav-st-radius;
       color: $grav-co-interactive-secondary-contrast;
     }
 
@@ -28,6 +29,7 @@
   &:not([href]) {
     background-color: $grav-co-interactive-primary;
     border-color: $grav-co-interactive-primary;
+    border-radius: $grav-st-radius;
     color: $grav-co-interactive-primary-contrast;
   }
 }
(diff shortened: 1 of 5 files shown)


src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss
```diff
@@ -10,6 +10,7 @@
     &:visited {
       background-color: $grav-co-interactive-secondary;
       border-color: $grav-co-interactive-secondary;
+      border-radius: $grav-st-radius;
```
c1rrus: Not sure if the rounding makes sense on the block links. They get used in the nav bar and since the bottom border is used to create a pseudo underline effect, it looks a little odd when the edges curve up.

src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss
```diff
@@ -28,6 +29,7 @@
   &:not([href]) {
     background-color: $grav-co-interactive-primary;
     border-color: $grav-co-interactive-primary;
+    border-radius: $grav-st-radius;
```
c1rrus: Same comment as above

james-nash:
@ThePeach @franz899 Should we perhaps close this PR without merging?

It's not needed to support the MVP website. Also our look and feel has evolved a bit since this was done - we have hero animations, flourishes, logos, etc., so if we were going to make any additional alterations to the Gravity look and feel, I think they need to be considered in its wider context.

franz899:
I agree, we can come back to it later.

PR #50, opened 2018-03-21
Footer social media links

patrickbower:
Closes WEB-126

To see changes see footer ie: `/?p=pages-homepage`


src/styleguide/_patterns/03-organisms/00-page-structure/01-page-footer.mustache
```diff
@@ -1,3 +1,12 @@
 <footer role="contentinfo">
   <p>{{footer.copy}} <a href="{{{url}}}">{{> atoms-logotype(label: "buildit @ wipro digital") }}</a></p>
-</footer>
\ No newline at end of file
+  <ul class="grav-c-list-horizontal">
+    {{#footer.social}}
+      <li>
+        <a href="{{url}}">
+          {{> atoms-inline-svg-symbol:grav-c-icon(symbolId: '{{icon}}', width: 100, height: 100 label: {{label}}) }}
```
james-nash: Why not just include `atoms-icon`? That would save you needing to manually set the class name and also make Pattern Lab display a link to that atom in the footer's "pattern info" panel.

ThePeach:
Just a personal preference, but I would have doubled the size of the icons. 
Maybe something we need to be aware of during the user testing session.

ThePeach:
All looks good. There's only two small things I'd like you to address which is about the naming conventions (which we might want to put into writing in case they are not clear/explicit):

1. The sass filename should be prefixed with an `_`, this will make sure it's not a "main" file, rather a partial.
2. You need to align the names you used for the sass, its file, and the representative mustache. As I've seen so far we used to follow this pattern: `grav-c-my-component` which will be in a file named `my-component` and the mustache file will use the same file name. The numerical prefix is transparent and used only by patternlab, so feel free to use them for ordering.

Aside from that all looks good!

PR #52, opened 2018-03-22
updated homepage copy layout

ThePeach:
- added has-centered-text utility-class
- added atom list-two-columns
- updated homepage layout and organisation


src/ui-lib/sass/06-utilities/_has-centered-text.scss
```diff
@@ -0,0 +1,3 @@
+.grav-u-has-centered-text {
```
franz899: This is an interesting name for a css class? Isn't this naming convention usually used for method that return a boolean?

src/ui-lib/sass/gravity.scss
```diff
@@ -81,3 +82,4 @@
 // === Utilities layer ===
 // Utilities and helper classes with ability to override anything which goes before in the
 // triangle, eg. hide helper class.
+@import '06-utilities/has-centered-text';
```
franz899: Should we have utility classes grouped by functionality? This one is a very specific file name.

ThePeach: fair point, I'll get it a generic `text` file for now

src/styleguide/_patterns/03-organisms/00-page-structure/03-section.mustache
```diff
@@ -7,12 +7,12 @@
     {{#blockquote}}<blockquote>{{{blockquote}}}</blockquote>{{/blockquote}}
     {{#inlineRow}}{{>organisms-list-inline-row}}{{/inlineRow}}
     {{#imgCards}}{{>organisms-list-img-cards}}{{/imgCards}}
-    {{#orderedList}}
-    <ul>
+    {{#twoColumnsList}}
```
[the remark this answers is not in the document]
ThePeach: it wasn't an ordered list in the first place 🤦‍♂️  it made more sense to have a more generic variable name that controls it.

PR #60, opened 2018-03-27
Web 125 - Animated burger menu

wipro-zara:
Closes ticket 125:

Creation of an animated burger menu that turns into an "x" on click.
Also has animated focus/hover state where lines in menu icon move further apart.

Initially had a border that would change colour on focus/hover, looks cleaner without one. Have a look and let me know if you agree.


src/ui-lib/sass/05-components/01-atoms/06-buttons/_08-toggle-menu.scss
```diff
@@ -1,31 +1,105 @@
...
+    &[aria-pressed='true']:hover,
+    &[aria-pressed='true']:active,
+    &[aria-pressed='true']:focus {
+      background-color: inherit;
+    }
+  }
+  
+  &:hover,
+  &:active,
+  &:focus {
```
james-nash: Don't need this block either :-)

(No need to set the cursor since the `<button>` element already make browsers show a suitable cursor when hovering over them)

src/ui-lib/sass/05-components/01-atoms/06-buttons/_08-toggle-menu.scss
```diff
@@ -1,31 +1,105 @@
...
+  &:hover &__icon::before {
+    transform: translateY(-$grav-tr-icon-hover);
+  }
+
+  &:hover &__icon::after {
+    transform: translateY($grav-tr-icon-hover);
   }
 
   &[aria-pressed='true'] {
     background-color: inherit;
```
james-nash: Can remove this background property too :-)

src/ui-lib/sass/05-components/01-atoms/06-buttons/_08-toggle-menu.scss
```diff
@@ -1,31 +1,105 @@
...
-    color: $grav-co-bg;
+    transition: all $grav-tr-speed-fast;
+    
+    &:hover,
+    &:active,
```
james-nash: ...and this block can go too.

src/ui-lib/sass/05-components/01-atoms/06-buttons/_08-toggle-menu.scss
```diff
@@ -1,31 +1,105 @@
...
+  }
+
+  &:hover &__icon::after {
+    transform: translateY($grav-tr-icon-hover);
   }
 
   &[aria-pressed='true'] {
     background-color: inherit;
-    color: $grav-co-bg;
+    transition: all $grav-tr-speed-fast;
```
james-nash: Don't need the `transition` property on the button itself, since nothing changes. You're only doing transitions on the icon elements within it.

src/styleguide/_patterns/02-molecules/00-navigation/00-nav-menu.mustache
```diff
@@ -1,5 +1,5 @@
 <button type="button" class="grav-c-toggle-menu" aria-pressed="{{control.pressed}}">
-  {{control.label}}
+  <span class="grav-c-toggle-menu__icon"></span>
```
ThePeach: would it be useful to move this tag into a separate atom? 
in which case you have two solutions: 
1. create a normal pattern and show it in the list of atoms/icons in patternlab
2. create a hidden pattern (prefixing it with `_` underscore)

src/ui-lib/sass/00-settings/_animations.scss
```diff
@@ -2,3 +2,4 @@
    Animation Variables
 */
 $grav-tr-speed: 250ms;
+$grav-tr-speed-fast: 400ms;
```
ThePeach: 400ms > 250ms, i.e. it's _slow_ rather than fast.

src/ui-lib/sass/05-components/01-atoms/06-buttons/_08-toggle-menu.scss
```diff
@@ -1,31 +1,105 @@
...
+$grav-tr-icon-rotate: 135deg;
+$grav-tr-icon-scale: 1.2;
+$grav-size-icon: 33px;
+
 .grav-c-toggle-menu {
+  display: flex;
+  height: $grav-size-icon;
   margin-top: 0;
-  border: $grav-st-thickness solid $grav-co-neutral-asphalt;
+  border: hidden;
```
ThePeach: that's only `border-style` that can be `hidden`, remember that `border` is just a shorthand property.

ThePeach:
It would be nice to try to have a very fast ease-out animation when the menu expands... if it's not too much work. We can think of doing this in a separate ticket if that's OK with you.

src/ui-lib/sass/05-components/01-atoms/06-buttons/_08-toggle-menu.scss
```diff
@@ -1,31 +1,97 @@
...
+    }
+     
+    &[aria-pressed='true']:hover,
+    &[aria-pressed='true']:active,
+    &[aria-pressed='true']:focus {
+      background-color: inherit;
+    }
+  }
+  
+  &,
```
ThePeach: even if we're replicating the code, wouldn't it be better to have this property moved at the top together with the rest of the other properties? in this way you can see all at once what applies to what.

PR #62, opened 2018-03-28
Eyeglass support

franz899:


package.json
```diff
@@ -74,7 +74,9 @@
     "stylelint-scss": "^2.2.0"
   },
   "dependencies": {
-    "@buildit/gravy": "^2.0.1"
+    "@buildit/gravy": "^2.0.1",
```
james-nash: Oh! Good spot! :-)

eyeglass-exports.js
```diff
@@ -0,0 +1,8 @@
+const path = require('path');
```
c1rrus: Just realised something: The `srcSassDir` returned by `index.js` is already an absolute path. Using `path` to join it with `__dirname` is therefore redundant here and can be removed. Sorry I didn't spot this earlier!

franz899: 👌

PR #68, opened 2018-04-04
WEB-141 - Animated hero dots image

jimmyj-wipro:
This is the initial work for the hero banner for the js work that will be added to the buildit website
- _hero.scss : changed to gradient, added canvas css
- 02-hero.mustache : removed bg url, closed p tag

### DIFF
--- a/src/styleguide/_patterns/03-organisms/00-page-structure/02-hero.mustache
+++ b/src/styleguide/_patterns/03-organisms/00-page-structure/02-hero.mustache
@@ -1,4 +1,5 @@
 <section class="grav-c-hero">
+  <canvas></canvas>
   {{#hero.content}}
     {{#paragraph}}<p>{{{paragraph}}}</p>{{/paragraph}}
     {{#pullout}}<blockquote class="grav-c-pullout">{{{pullout}}}</blockquote>{{/pullout}}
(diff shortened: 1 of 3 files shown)


src/styleguide/_patterns/03-organisms/00-page-structure/02-hero.mustache
```diff
@@ -1,6 +1,6 @@
-<section class="grav-c-hero" style="background-image: url('{{hero.background}}')">
+<section class="grav-c-hero">
   {{#hero.content}}
```
james-nash: We should add the `<canvas>` element here (even though it's dormant without any CSS), as it's a useful reference to developers viewing the markup for the hero organism in our style guide.

src/ui-lib/sass/05-components/03-organisms/_hero.scss
```diff
@@ -1,21 +1,23 @@
...
   > * + * {
     margin-top: $grav-sp-xl;
   }
+
+  > canvas {
+    max-width: initial;
+    position: absolute;
+    top: 0;
+    left: 0;
+    margin: 0;
```
james-nash: Need to add `width: 100%; height: 100%;` to ensure that the canvas always covers the entire hero section (which has an unpredictable width & height).

PR #70, opened 2018-04-06
Reduces max number of word per row on high resolutions

franz899:


src/ui-lib/sass/03-elements/_section.scss
```diff
@@ -1,6 +1,23 @@
 section {
   margin-top: 0;
 
+  p {
```
james-nash: Hmmm... the code below is mostly fine. However, I'd like the `max-width` value to be extracted as a variable and put under `settings/` somewhere. Likewise, the `margin-left` value ought to use one of the ones defined in `_spacing.scss`.

_However_, I wonder if a better solution is simply to adjust the max-width of our page content container. That way everything will be constrained to a sensible line-length. We can avoid needing to indent the heading by an arbitrary amount then too.

Having poked around the code, I think it may be worth revisiting that and the breakpoints. They're already being used somewhat inconsistently. Let's discuss with the wider team before proceeding.

franz899: I'm ok to move the value into the typography css as `$grav-line-max-width`. I didn't assign `max-width` to the container because that would make everything narrower, even on designs where we don't want to use it, like locations or careers.

By applying a fixed margin left to the `hx` I ensured that the paragraphs are always going to be correctly aligned and that there is not going to be too much distance between a section title and its content.

ThePeach:
I tend to agree with @franz899 about the solution, but I have my problems here, as I think of this as the approach we are going to use for generic content, especially when we will integrate the blog.

IF we agree that our content copy is going to be mostly organised using `<section>`s and  `<article>`s, then we need to address _at least_ on `<h*>`s and `<p>`s.

src/ui-lib/sass/03-elements/_section.scss
franz899: Just a thought. What about a new class that is going to take care about settings these rules?

Wrapping headers, paragraphs and other tags in a div is not that bad, it's what they are made for. That div will only have styling properties.


Example:

src/ui-lib/sass/06-utilities/_text.scss
```diff
@@ -1,3 +1,9 @@
 .grav-u-text-centered {
   text-align: center;
 }
+
+.grav-u-max-word-count-m {
+  max-width: gravy-breakpoint(regular);
```
ThePeach: I think this ought to be in `em` rather than `px`, otherwise it defies the purpose of this piece of work.
Aside from this, I wouldn't even use the breakpoint as this can change in the future, and I'd rather introduce a new variable.

franz899: I did use a new variable and then changed it. It was 40em before. I'll revert it to that.

PR #74, opened 2018-04-09
Web-157 (uneven gap between lines in burger menu)

wipro-zara:
Closes ticket WEB-157 
 -changed units from rem to px - as rem's decimal precision seemed to be rounding incorrectly in some browsers, causing the gap offset.
-added new local variable for icon width to keep all sizes in px.

-fixed lint error


wipro-zara:
@jimmyj-wipro @ana-buildt 
The issue with rem is that the sizing is relative, so in different browsers on a certain computers/devices the calculated rounded value may vary - which is what I think caused the initial bug to begin with. And the bug only occurred in chrome.

I'd rather not use pixels as it's not the best practice, but since it's an absolute value, I think in this specific scenario it would be the best choice.

PR #77, opened 2018-04-12
Aligned baselines of location and jobs in job-list.

james-nash:
The goal was to make the baseline of the H2 text (the location of the jobs) and the first row of job card text to the right be aligned.

After some trial and error, I found that switching to flexbox to do the layout was the only way the `align-items: baseline;` had the desired effect consistently in Chrome, Firefox and Safari.

Maintaining the CSS grid-based layout didn't work everywhere. Only having flex on the outer `job-list` component but continuing to use CSS columns on the job cards didn't work in Firefox. etc.

### DIFF
--- a/src/ui-lib/sass/05-components/03-organisms/_job-list.scss
+++ b/src/ui-lib/sass/05-components/03-organisms/_job-list.scss
@@ -11,6 +11,7 @@
   > header {
     width: 30%;
     margin-right: $grav-sp-m;
+    flex-shrink: 0;
   }
 
   > ul {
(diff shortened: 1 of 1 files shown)


src/ui-lib/sass/05-components/03-organisms/_job-list.scss
```diff
@@ -4,30 +4,44 @@
...
-    grid-template-columns: 30% 1fr;
+    display: flex;
+    align-items: baseline;
   }
 
-  > header > p {
-    @include typi(small-text);
-    margin-top: 0;
+  > header {
+    width: 30%;
```
danbull: The width of the `header` isn't rendering as a consistent width which means things aren't aligned as well as they could be.

image

james-nash:
@danbull Good spot! I've just pushed a fix for that issue

PR #81, opened 2018-04-16
Web 162 - Backround Banner to Logos

wipro-zara:
Closes ticket WEB-162

The background banner to the logos has been switched from the **list-image-links** itself over to a background-banner modifier of the **container**. 

Stylings adjusted accordingly as well as any code debris clean up within *list-image-links** 

Modified homepage mustache & json files to support above changes.

Added BP logo and adjusted the canvas vertical sizes of all other logos for aesthetics.


src/styleguide/_patterns/04-templates/00-homepage.mustache
```diff
@@ -4,5 +4,11 @@
   {{#sections}}
   {{>organisms-section}}
   {{/sections}}
+  <section>
+    <p class="grav-u-text-centered">Here are some of the clients we have worked with:</p>
+    <div class="grav-o-container-background-banner">
```
ThePeach: there is a bigger piece of work that I think it would be worth to address:

screen shot 2018-04-16 at 16 24 59

By looking at the above screen, the previous paragraph seems to be belonging to the previous section, while it's not. You're basically using colour to group sections, so I'd suggest a couple of changes here which should also simplify the whole thing:

1. remove the wrapping div for the `list-image-links`, this is not needed anymore as it was providing the background
2. change the `grav-o-container-background-banner` to be something like a style modifier for the section that would set the background to it
3. while doing the above you might need to adjust margins and paddings to better separate the current section to the previous one
4. switching back into using the `organism-section` will then be quite straightforward, so if you do so, remember to update the examples for the organism and its description/documentation perhaps.

src/styleguide/_patterns/05-pages/00-homepage.json
```diff
@@ -36,31 +36,21 @@
           },
           {
             "paragraph": "To learn more, check out our <a href='https://medium.com/buildit/'>blog</a>."
+          },
+          {
+            "paragraph": "Here are some of the clients we have worked with:"
```
ThePeach: as I was trying to explain previously, this should belong to the following section, as it is now it does not make any sense from a semantic point of view.

src/ui-lib/sass/03-elements/_section.scss
```diff
@@ -4,4 +4,9 @@ section {
   + section {
     margin-top: $grav-sp-xl;
   }
+  
 }
+.section--background-banner {
```
ThePeach: this should go in a more appropriate place, as by the look of it, it can be applied to anything. Perhaps as a utility class?

ThePeach:
also, don't forget to fix the conflicts (in this particular instance I'd suggest to merge from master, remove the file and re-create it from scratch.

src/ui-lib/sass/04-objects/_container-banner.scss
```diff
@@ -0,0 +1,4 @@
+.grav-o-container-banner {
+  padding: $grav-sp-xl $grav-sp-m $grav-sp-m;
```
ThePeach: perhaps you want to create a new `$grav-sp-inset-squished-xl` and use that one?

ThePeach: You also want to showcase this class next to the `organism-section` and write some documentation for it, this seems to be describing an alternative type of container compared to `grav-o-container` so you might want to make sure this is well understood.

src/ui-lib/sass/05-components/03-organisms/_list-image-links.scss
```diff
@@ -1,8 +1,9 @@
 .grav-c-list-image-links {
+  $image-width-m: calc(100% * (1/6) - #{$grav-sp-m});
+  $image-width-l: calc(100% * (1/6) - #{$grav-sp-l});
 
-  padding: $grav-sp-l $grav-sp-m calc(#{$grav-sp-l} - #{$grav-sp-m});
-  background-color: $grav-co-neutral-ashtray;
-  column-gap: $grav-sp-l;
+  padding: 0 $grav-sp-xl $grav-sp-l;
```
ThePeach: If you address the padding of the container as explained above, you don't even need to set the padding on this element, and let the margin-top deal with the vertical-spacing (which is already set and you don't need to do anything). The idea being is that every element should be context-free, and the container should dictate how things are set. There are few exceptions to this, but this is the overall ideal.

src/ui-lib/sass/04-objects/_container-banner.scss
wipro-zara: inset-squished is 2 dimensional, so its top/bottom & right/left. And what I'm doing is setting top as XL, right/left as M, and bottom as M..so I don't think adding a new one is necessary here? Correct me if I'm wrong.

ThePeach: no but what I'm suggesting is to have more or less the same padding that the other container is using, that is using same vertical/horizontal padding (maybe xl/l ?)

ThePeach: In that way you also will help the spacing for the icon list (see my other comment)

ThePeach:
success

PR #88, opened 2018-04-20
Example of navbar with active page bold text

franz899:


src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -5,6 +5,16 @@
 header[role=banner] {
   background-color: $grav-co-neutral-charcoal;
 
+  a {
```
james-nash: Can you move these changes to the `src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss` file instead. It's the individual nav link components that we're chaning here, not the list itself.

PR #91, opened 2018-04-25
SVG cleanup

james-nash:
Miscellaneous SVG improvements:

### Fixing validation errors
The old Buildit logo SVG had some inline style-related attributes (most likely things added by Inkscape) which were not valid SVG:
image

To remove these attributes (which we don't need anyway, since our inlined SVGs are styled via the page's CSS) and prevent the problem from recurring in the future, the relevant `gulp` task has been updated. It now strips off such attributes.

### Updating the Buildit logo
The Buildit logo was recently updated on the website (for the favicons) but that change hadn't worked its way upstream into Gravity. This has now been fixed.

To allow for more flexible usage, an additional "hollow" variant has been added as well. This is because the new logo has a circular outline. Depending on the context this may be invisible if it matches the colour of the surrounding background. In other situations, we intentionally want to display it though.

### Title IDs for better a11y
So far, we had assumed that instances where our SVG symbols are _displayed_ would automatically reference the `<title>` element of the SVG symbol being referenced by `<use>`. A11y testing on the website has determined that this is not the case. The work-around is to add a `<title>` into the inline SVG or to give it an `aria-label` attribute. This is what the website is doing at the moment.

Unfortunately, that does mean title text may be duplicated on the page if the same symbol is displayed more than once. Furthermore, it's harder to maintain if the title text ever needs to be changed since every instance needs to be updated.

To prevent this in future, every SVG symbol's `<title>` will automatically have an ID attribute added to it by the `gulp` build. Inline SVGs can then reference that ID via `aria-labelledby`.


james-nash:
@franz899 As discussed, that sounds good. We can certainly expose the SVGO config options to consumers via Gravity's `index.js`.

Thinking about it a bit more, exposing the whole Gulp pipe with SVGO and Cheerio probably isn't ideal. I'd prefer to keep them as `devDependencies`, since it's likely that future consumers of this library won't necessarily be using Gulp (e.g. Buildit's Twig app uses Angular's `ng build`).

That being said, we _could_ extract this (and other?) Gulp stuff into a separate npm package than Gravity and the website can then both use.

In any case, I propose that whatever we do for this, we do it in a separate PR from this one. Sound good?

PR #96, opened 2018-04-25
Bug Fix (WEB-204): Changed inline text link colour on active state

wipro-zara:
Changed inline text link colour on active state. 

Move active below visited for specificity - so that active overrides the visited colour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - `<header role="banner">` page title now has a hover state across all pages
+
 ### Changed
 - `a[href]` has new contrast appropriate active state text colour in `inline-text`. Set to override visited state.
+
 ### Fixed
 - `<a>` elements now are no longer smaller than their child elements in `list-inline-row` by being set to inline-block.
+
 ### Removed
 - unecessary background for `<a>` element in active state 
+
+
 ## [0.8.0] - 2018-04-25
 ### Added
 - Hollow version of Buildit logo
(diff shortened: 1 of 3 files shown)


src/ui-lib/sass/03-elements/_inline-text.scss
```diff
@@ -9,14 +9,15 @@ a[href] {
...
-    color: inherit;
-  }
-
   &:visited {
     color: $grav-co-interactive-secondary-visited;
   }
+
+  &:active {
+    background: $grav-co-interactive-secondary;
+    color: $grav-co-neutral-dirty-snow-white;
```
james-nash: Can you swap this for `$grav-co-interactive-secondary-contrast` please (it's defined in `_colors.scss`). It ultimately still ends up being Dirty Snow White, but we want to try to avoid directly referencing the named colours in our CSS.

The idea is that using intermediate variables, whose names express a purpose, mean we could reskin/rebrand a lot of our UI simply by fiddling with whatever named colours they map to. Otherwise, it becomes a tedious "find and replace" job.

ThePeach:
aside from making the colour explicit, which is fine, I don't understand why we have such a dark on dark active state... @james-nash ? maybe this is stuff that gets fixed with the other ticket about colours consistency?

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 
+
+### Fixed
+- `a[href]` has new contrast appropriate active state text colour in `inline-text`. Set to override visited state.
+- removed unecessary background for `<a>` element in active state
```
ThePeach: this should be under "removed" heading

PR #100, opened 2018-04-26
Bug Fix (WEB-214) Target IE10+ for tools > container

ana-buildt:
This is [a IE11 bug](https://digitalrig.atlassian.net/browse/WEB-214) that happens when using margin: auto to fill all the available space between flex items. Here's more info [about this bug](https://github.com/philipwalton/flexbugs#flexbug-15)

### DIFF
--- a/src/ui-lib/sass/01-tools/_container.scss
+++ b/src/ui-lib/sass/01-tools/_container.scss
@@ -2,6 +2,9 @@
 @mixin grav-l-container($no-margin: false) {
   max-width: $grav-page-content-max-width;
 
+  // IE10+ bug that happens when using margin: auto
+  // to fill all the available space between flex items.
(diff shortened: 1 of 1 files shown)


src/ui-lib/sass/01-tools/_container.scss
```diff
@@ -2,6 +2,10 @@
 @mixin grav-l-container($no-margin: false) {
   max-width: $grav-page-content-max-width;
```
ThePeach: please add this link in a comment on why this is a bug: https://alastairc.ac/2014/08/flexbox-ie11-bugs/ and add also the link to OUR bug report

wipro-zara:
I like it! It works perfectly, just read through some documentation for it and it seems like the only solution. Nice on!

PR #107, opened 2018-05-01
WEB-216 Fixed spacing between flex items when wrapped

wipro-zara:
Closes ticket WEB-216.
Added negative margins so items have improved spacing between another when wrapped.


james-nash:
...but please also update the CHANGELOG.

PR #109, opened 2018-05-01
FireFox image responsiveness fix

### LINKED ISSUE #108: Navigation bar is not fully responsive in Firefox
As it currently stands, the navigation menu in the header will crop out the last item (blog) when scaled down between 700-760px. This issue has only been found in **Firefox**.

**To reproduce bug, follow these steps:**
1. Start either a Firefox browser
2. Paste "https://www-beta.buildit.digital/" into the address bar
3. Minimize and scale the browser down to just before the navigation turns into a burger menu (between 700px-760px)

**Expected outcome:**
The navigation menu remains responsive and each item scaled down accordingly

**Actual outcome:**
The last item on the navigation menu (blog) is cropped out.
Currently only occurs in Firefox

**UPDATE**
Also occurs in **IE11** & **Edge**.

wipro-zara:
-set width to 100%

Fixes #108


src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -56,12 +56,7 @@ body > header {
     display: block;
     max-height: 33px; // Otherwise IE11 makes it 150px tall :-(
     fill: $grav-co-bg;
-
-    // Bump up logo size
-    @media (min-width: gravy-breakpoint(medium)) {
-      height: 2.5rem;
```
james-nash: The current solution makes things a bit wonky on smaller screens - the SVG _container_ fills 100% of the space right of the hamburger button, but the logo is centered within that, so it is no longer flush with the left edge of the header:

image

To resolve this, re-instate the media query you removed, and move the `width: 100%;` within it (replacing the `height: 2.5.rem` that was there before).

james-nash: So, all together, it should look like this:
```sass
  .grav-c-logo {
    display: block;
    max-height: 33px; // Otherwise IE11 makes it 150px tall :-(
    fill: $grav-co-bg;

    // Bump up logo size
    @media (min-width: gravy-breakpoint(medium)) {
      width: 100%;
      max-height: none;
    }
  }
```

james-nash:
@wipro-zara One more thing: Please update the CHANGELOG as well. Thanks!

PR #114, opened 2018-05-02
Changing bg of hero

danbull:

### Overall Approach

#### JSON

#### Patterns/templates

#### SASS

#### Other (images)


src/styleguide/_patterns/03-organisms/00-page-structure/02-hero.mustache
```diff
@@ -2,5 +2,5 @@
   <canvas></canvas>
   {{#hero.pullout}}<blockquote class="grav-c-pullout">{{{hero.pullout}}}</blockquote>{{/hero.pullout}}
```
danbull: @james-nash should this maybe be changed from a `blockquote` to a heading?

ThePeach: I think we are in agreement this is semantically correct.

james-nash: @danbull Nah, I think blockquote is appropriate here.
For the website homepage, the "buildit @ wipro digital" logo in the page header is the closest thing we have to a heading. #110 is covering that.

danbull: Ok but since it will often contain text that doesn't have an original source it seems a bit weird to me. I'd be interested what it means from an SEO perspective.

james-nash: Hmm... that's a good point. Maybe a simple `<p>` would suffice.

Regardless, I propose we leave it as is for now so as not to delay the site launch. I'm pretty sure that besides headings, the effect of semantics on SEO is relatively low (I think search engines sift through so much garbage markup that they've almost given up on tags being used as intended).

Perhaps add an issue or card to [Gravity's ideas backlog](https://github.com/orgs/buildit/projects/4#column-2645503) for us to revisit this at a later date?

PR #122, opened 2018-05-08
fixes #117 - missing patterns

### LINKED ISSUE #117: Missing patterns cause output in terminal
h2. steps to reproduce

run `npm start`

h2. expected results

no errors/warnings in the terminal

h2. actual results

in the output you will see the following:

```
Could not find pattern referenced with partial syntax organisms-list-inline-row. This can occur when a pattern was renamed, moved, or no longer existsbut it still called within a different template somewhere.
Could not find pattern referenced with partial syntax organisms-list-img-cards. This can occur when a pattern was renamed, moved, or no longer exists but it still called within a different template somewhere.
Could not find pattern referenced with partial syntax organisms-list-image-links. This can occur when a pattern was renamed, moved, or no longer exists but it still called within a different template somewhere.
```

ThePeach:
Ticked only the ones that applies. This addresses a problem with PatternLab itself.

### Overall Approach

#### JSON

#### Patterns/templates

#### SASS

- [ ] The SASS (and related pattern files) follow the [naming conventions](/docs/naming-conventions.md), architecture and structure outlined in the documentation.

#### Other (images)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- fixed spurious info output in terminal by moving all list patterns under `molecules/lists`
+
 
 ## [0.9.1] - 2018-05-03
 ### Added


ThePeach:
@james-nash had a hunch that you couldn't include organisms into other organisms (from a PL perspective it doesn't make sense) and apparently it made sense. This doesn't explain the problem itself as it should allow for instance inclusion of molecules into other molecules... Will raise in PL chan and see if there's an explanation

PR #126, opened 2018-05-09
Fixed fake spaces in CHANGELOG that were screwing up the rendering

c1rrus:
Some of our older CHANGELOG entries weren't rendering correctly due to non-space characters being present (can happen in VS Code is you accidentally type SHIFT + SPACE). This fixes that.

Most (all?) of the checklist below does not apply. Will only tick things that are relevant.

### Overall Approach

#### JSON

#### Patterns/templates

#### SASS

- [ ] The SASS (and related pattern files) follow the [naming conventions](/docs/naming-conventions.md), architecture and structure outlined in the documentation.

#### Other (images)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Stylelint issues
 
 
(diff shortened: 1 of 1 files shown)


james-nash:
Does it? Does it _really_?

🤣

PR #134, opened 2018-05-15
Fix footer ie

danbull:

### Overall Approach

#### JSON

#### Patterns/templates

#### SASS

#### Other (images)

Fixes part of https://digitalrig.atlassian.net/secure/RapidBoard.jspa?rapidView=229&projectKey=WEB&modal=detail&selectedIssue=WEB-259

### DIFF
--- a/src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
+++ b/src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
@@ -25,15 +25,15 @@ body > footer {
   }
 
   svg {
-    display: block;
+    display: inline-block;
     margin: 0 auto;
     width: auto;
     height: 1em;
-    max-width: 7rem;
   }
 
   .grav-c-logo {
     border-bottom: 1px solid $grav-co-neutral-dirty-snow-white;
+    max-width: 7rem;
 
     &:hover {
       border-bottom-color: $grav-co-neutral-ashtray;
--- a/src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
+++ b/src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
@@ -7,7 +7,6 @@ body > footer {
 
 
   > * {
-    @include grav-l-container;
     padding-bottom: $grav-sp-s;
     text-align: center;
 
--- a/src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
+++ b/src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
@@ -26,9 +26,10 @@ body > footer {
 
   svg {
     display: block;
+    margin: 0 auto;
     width: auto;
     height: 1em;
-    margin: 0 auto;
+    max-width: 7rem;
   }
 
   .grav-c-logo {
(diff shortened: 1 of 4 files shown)


danbull:
By default SVGs in IE render at 150hx300w and `width: auto;` does not override this. Unfortunately the only easy way to fix this is to be a bit more explicit.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
```diff
@@ -7,7 +7,6 @@ body > footer {
 
 
   > * {
-    @include grav-l-container;
```
james-nash: We shouldn't be removing this. This is what ensures that footer content does not exceed the max content with that the rest of the page has.

danbull:
@ThePeach I thought that myself, I'm going to go ahead and change that.

james-nash:
That `300 x 150` size is actually a standard size for "replaced elements" in HTML, of which inlined SVGs are one. I wrote it up in the pattern info for http://style.buildit.digital/?p=atoms-inline-svg-symbol

Note that, it might not always _look_ like it, since its the bounding box of the SVG that gets sized to 300 x 150px. By default the actual graphic will be centered within that bounding box.

In order to "shrinkwrap" an inlined SVG's bounding box to the actual graphic, you need to give it `width` and `height` attributes. You can see an example of this with: http://style.buildit.digital/?p=atoms-logotype - note how 300 x 33 is being set in the HTML itself!

With those attributes in place, the inline SVG behaves much like an ``. If you alter it's width in CSS (while leaving the `height` as `auto`) the actual graphic scales _proportionally_.

That's why our http://style.buildit.digital/?p=atoms-icon component sets its intrinsic size to 32 x 32, because it's assuming that all icons have a square aspect ratio.

Assuming we want icons to appear at a different size to 32px x 32px, all we need to do is set a width in CSS - which could be any unit you like `em`, `rem`, `%` etc.

Looking at the existing footer CSS code, that message got lost and all kinds of weird width and height setting is going on. :-(

I'd try trimming it all away and simply setting the `width` to whatever the desired value is. Hopefully that'll also work in IE.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
```diff
@@ -26,21 +25,23 @@ body > footer {
   }
 
   svg {
-    display: block;
+    display: inline-block;
+    margin: 0 auto;
     width: auto;
```
james-nash: I think most of these properties are intended for the Buildit logo, not the icons.

Maybe change the selector to `.grav-c-logo`. That way, you don't need to undo all that stuff for the icons further down.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
```diff
@@ -26,21 +25,23 @@ body > footer {
...
     border-bottom: 1px solid $grav-co-neutral-dirty-snow-white;
+    max-width: 7rem;
 
     &:hover {
       border-bottom-color: $grav-co-neutral-ashtray;
     }
   }
 
   .grav-c-icon {
+    max-width: 1.5em;
```
james-nash: As per my big comment - you _should_ be able to replace setting `max-width` and `height` here with simply setting `width: 1.5em;`.

PR #135, opened 2018-05-17
Made header nav menu more flexible

danbull:

### Overall Approach

#### JSON

#### Patterns/templates

#### SASS

#### Other (images)

Fixes https://digitalrig.atlassian.net/browse/WEB-181


src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -14,16 +14,16 @@ body > header {
...
     }
   }
 
   .logo-main {
     width: 70%; // 2/3 ish of the space
 
     @media (min-width: gravy-breakpoint(medium)) {
-      width: auto;
+      margin: $grav-sp-m $grav-sp-m 0 0;
+      max-width: 250px;
```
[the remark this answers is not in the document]
danbull: I thought about that but many people I've worked with go with the principle of YAGNI and if you're only using a variable once then it's a bit pointless having it https://sass-guidelin.es/#variables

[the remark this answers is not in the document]
james-nash: Fair enough.

src/ui-lib/sass/05-components/02-molecules/00-navigation/_00-nav-menu.scss
```diff
@@ -16,6 +16,7 @@
 
     @media (min-width: $trigger-breakpoint) {
       display: flex;
+      flex-wrap: wrap;
       margin-top: 0;
       text-align: right;
```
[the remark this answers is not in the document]
danbull: I gave the last child `margin-right:  0;` instead. Assume that's fine?

src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss
```diff
@@ -1,7 +1,6 @@
 .grav-c-block-link {
   display: inline-block;
-  padding: $grav-sp-inset-squished-xs;
-  padding-right: $grav-sp-m;
+  padding: $grav-sp-xs;
```
james-nash: Hmm... looking at this with a fresh pair of eyes, I've noticed a couple of things:

* This PR's code produces a larger visible gap between adjacent nav link texts than on the current Buildit site. This is because in this PR the gap is 2 * `$grav-ps-xs` padding + `$grav-sp-m` margin, whereas on the current Buildit site there is no margin between successive block links, so it's only the two lots of padding that create the space between words.
* The padding on block links is really a hangover from when they had the underline effect, but we've since removed that, so it's perhaps redundant now. A consequence of that is that the right-most link text is not flush with the boundary of the page's content area. Likewise, if the nav links wrap onto a new line, the left-most ones will also be inset by the padding.

I therefore propose the following:

* Give block links _only_ top and bottom padding (`$grav-sp-xs`). Arguably, this might be redundant too, but I feel it's worth keeping as it:
    * Makes the clickable/tappable link area a bit bigger and thus easier for users to hit (Fitts' Law)
    * Helps keep the overall layout of the nav links and header consistent with what's currently live on the Buildit website

This would achieve the following:

* The margin between adjacent block links in a nav menu that this PR introduces is equivalent to the two paddings in the current site, so visually the nav menu should look the same.
* Any links that are at the left or right edges of the page will have their link text flush with the edge of the content area, which I think will look a bit neater.

src/ui-lib/sass/05-components/02-molecules/00-navigation/_00-nav-menu.scss
```diff
@@ -38,13 +38,15 @@
     }
   }
 
+  li:last-child {
```
james-nash: Using a negative right margin on the `ul` will be a more robust solution.

As it stands, if the nav links wrap over multiple lines, then only the last link in the last row will have its right margin removed. The right-most links in preceding rows will still have a right margin.

If you use a negative margin on the `ul`, then it will work for _all_ rows of links.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -14,16 +14,16 @@ body > header {
...
+      padding-top: 0;
     }
   }
 
   .logo-main {
     width: 70%; // 2/3 ish of the space
 
     @media (min-width: gravy-breakpoint(medium)) {
-      width: auto;
+      max-width: 250px;
```
james-nash: Perhaps we can bump up the `max-width` a little so that the logos size is closer to what's currently live on the website. I haven't measured it, but to my eyes the logo _looks_ a bit bigger on the website than it does in the styleguide with this PR.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -72,12 +70,11 @@ body > header {
     // Overflow, so that right edge of navlinks is flush
     // with screen edge
     width: calc(100% + #{$grav-page-content-inset});
```
james-nash: This oversized width is no longer needed since we don't have those borders anymore. This should have actually been removed a while ago - it's nothing to do with this PR. However, since we're here, could you please clean it up. Thanks!

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
danbull: So I reduced because right now there's not enough space for the Buildit website nav at ~700px (once the Medium logo is in place). With this new code the nav would wrap below the logo (better than hanging off the screen) but I don't think it's ideal. I can revert it though if you prefer?

src/ui-lib/sass/05-components/01-atoms/10-navigation/_00-block-link.scss
danbull: I only just noticed this comment but I think I've addressed the issues as I also noticed this today. The only main difference is I kept the left/right padding and made them flush with negative margin on the parent. Seems to work well?

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
james-nash: Ah Ok. That's fine then. Agree that we want to avoid wrapping on the Buildit website if we can.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed spurious info output in terminal by moving all list patterns under `molecules/lists`
 - Removed unnecessary universal selector inside hero component, which caused layout issues in IE
 - Fixed SVGs rendering at the wrong size in IE
+- Added a max width to the logo so there's more space for the primary nav
+- Primary nav will now wrap underneath the logo if there's not enough space for it
+
 
 ### Changed
 - Increased size and spacing of burger menu strokes
```
james-nash: Please add a note under "Changed" about renaming `grav-c-block-link` to `grav-c-nav-link` and call out that it's a _breaking change_.

src/ui-lib/sass/05-components/02-molecules/00-navigation/_00-nav-menu.scss
```diff
@@ -1,6 +1,7 @@
+$grav-c-nav-menu-link-margin: $grav-sp-s;
```
james-nash: Does this need to be a global SASS variable?

I think you could move it into the `.grav-c-nav-menu { }` block. If you do so, you don't need to use the `grav-...` prefix and could just call it `$nav-link-margin`. Thinking about it, perhaps `$nav-link-padding` is a more appropriate name, since that's what this represents (if I've understood the intent of your code correctly).

src/ui-lib/sass/05-components/02-molecules/00-navigation/_00-nav-menu.scss
```diff
@@ -16,17 +17,14 @@
 
     @media (min-width: $trigger-breakpoint) {
       display: flex;
-      margin-top: 0;
-      text-align: right;
+      flex-wrap: wrap;
+      margin: 0 (-$grav-c-nav-menu-link-margin) 0 (-$grav-sp-s);
```
james-nash: Perhaps use `$grav-c-nav-menu-link-margin` for the left margin value as well (or , `$nav-link-padding` if you rename it as suggested above). I think that would make the purpose of the negative margin clearer.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
ThePeach: please move these measures into a variable

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -58,26 +57,28 @@ body > header {
...
+      }
     }
   }
 
   .grav-c-nav-menu {
-    // Overflow, so that right edge of navlinks is flush
-    // with screen edge
-    width: calc(100% + #{$grav-page-content-inset});
-    margin-top: 0;
+    width: calc(100% + #{$grav-sp-s});
```
james-nash: If you change the nav menu as suggested, you can remove this line.

src/ui-lib/sass/05-components/02-molecules/00-navigation/_00-nav-menu.scss
james-nash: Also - I've just had an idea (sorry!) - we could move this out of the media query, so that the negative left & right margins always apply (thus always making the text of nav links flush with the edges of the nav menu).

Then, we could remove that oversized width in header's CSS which is basically trying to achieve the same thing.

PR #136, opened 2018-05-22
Updated Stylelint rules

james-nash:
This PR adds some stylelint rules that enforce [Gravity's naming conventions](https://github.com/buildit/gravity-ui-sass/blob/develop/docs/naming-conventions.md) for SASS variables and class selectors.

It correctly flagged up a few issues in our existing SASS code, of those...

* I have excluded 2 SASS vars from the linting as they are for configuring Gravy and Modularscale respectively and therefore cannot be renamed.
* I have fixed one remaining SASS var to conform to our conventions
* 2 class selector issues remain, but fixing them will require wider changes (breaking ones in one case), so I will raise separate issues for those rather than confusing this PR.

### Overall Approach

Not all of these are relevant to the changes in this PR, so only ticking ones that are.

#### JSON

#### Patterns/templates

#### SASS

#### Other (images)


james-nash:
Hmm.. looks like `npm ci` on Travis now checks dependencies for security vulnerabilities. The build fails because it's finding some. Will investigate later and see if I can resolve that. Please hold off reviewing this PR until then.

james-nash:
Ok, false alarm. It was just a timeout issue on Travis' end. I restarted the builds and they pass now. Removing the WIP label. **Review away!**

PR #145, opened 2018-06-20
CTA clean-up

### LINKED ISSUE #138: Remove `.is-disabled`, button and possibly secondary variants of CTA
Currently, the Buildit website _only_ uses the `grav-c-cta` class on `<a>` elements. However, we have a number of additional variations available in Gravity. I believe we can and should remove those to solve the following issues:

* The `is-disabled` class does not conform to [Gravity's naming conventions](https://github.com/buildit/gravity-ui-sass/blob/develop/docs/naming-conventions.md) (and, once PR #136 is merged, will cause linter errors) and is completely redundant. `button`s should use the native HTML `disabled` attribute and links cannot be disabled (they can be placeholders by omitting the `href` attribute, but that's not the same as being disabled and also only really makes sense in the context of nav links, which is not a valid use-case for CTA links anyway). **This class should therefore be removed completely.**
* Buttons and links are different. Buttons _do_ things and links _go_ somewhere. They also have different behaviour - buttons do not have context menus when you right-click them (or equivalent) whereas links generally do (e.g. to open in new tab, copy the URL, etc.). The use-case for CTA _links_ is clear and already in use on the Buildit website. I think the use-case for CTA _buttons_ is not clear though. We shouldn't be using buttons as fake links, so therefore they'd only be appearing in the context of forms or as controls in an app (e.g. on a toolbar) - either way, I'm struggling to see a strong need for a CTA variant of a button. **I therefore propose that we remove support for styling `button` elements with the `grav-c-cta` class.**
* In a similar vain, I struggle to see a strong need for a secondary variant of CTA links (or buttons). If the CTA's purpose is to be the one attention-grabbing thing on a page, why would we ever want a secondary variant? Would our normal link styling not suffice for that? **I therefore propose that we remove the `grav-c-cta--secondary` modifier class.** If an actual need for something like this arises in the future, we can always recreate it then.

james-nash:
Closes #138 

* Removed all redundant variants of the CTA atom. Now only the link one remains.
* Tidied up other link-related CSS to get rid of some redundant overrides
* Fixed a minor SASS formatting issue that was being flagged by the linter
* Updated some patterns to include the CTA pattern so that lineage links appear under pattern info

### Overall Approach

#### JSON

#### Patterns/templates

#### SASS

#### Other (images)


james-nash:
@danbull Muchas gracias! :-)

PR #181, opened 2019-01-30
Convert templates to Nunjucks

james-nash:
**Description**
* Configures Pattern Lab to use [Nunjucks](https://mozilla.github.io/nunjucks/) instead of Mustache
* Converts all Mustache templates to Nunjucks
* Refactors data to work better with Nunjucks pattern engine (which, for example, doesn't support `listitems.json` properly)

**Motivation**
Right now:
* Pattern Lab uses Mustache
    * The Mustache templates use features unique to Pattern Lab such as pattern params, styleModifiers and list items, which means they cannot easily be re-used outside of Pattern Lab
* The Buildit website has its own set of templates, written in Handlebars, which largely recreate the _same_ HTML structures we do within Pattern Lab
    * Furthermore, a number of Handlebars helpers have been configured because the site needed features that Handlebars didn't provide out of the box

We need to get to a state where both Pattern Lab and the Buildit website (and potentially more sites in future) share the same templates. This will cut down the maintenance burden and make it easier to roll out changes. Therefore, both need to align on the same template language.

Mustache is too basic for the site. Also, as mentioned above, our current Mustache templates are using proprietary Pattern Lab features anyway.

We could go with Handlebars, but it just so happens that Nunjucks can do out-of-the-box most of what the website needs but currently needs Handlebars helpers to achieve.

So, the decision was made to migrate to Nunjucks.

This PR is the first step. Once approved and merged, the next step will be to make builds export and publish the templates as a new NPM package, which the website and others can then consume.

**Notes**
All templates have been converted and our pattern library builds fine. The rendered HTML is the same as before.

But:
* In anticipation of the template files eventually being used outside of Pattern Lab, I have updated many of them to use Nunjucks' [macro](https://mozilla.github.io/nunjucks/templating.html#macro) feature. While it does add some complexity to the template, it enables other templates to `import` the macro (as opposed to `include` the template) and then pass values into the macro.
     * With includes, the included template references the same context data the parent template has. Therefore, when there are multiple instances of a template being included, it becomes difficult to provide different data to each one.
     * The macro mechanism on the other hand is akin to passing props to React components - you explicitly pass data _into_ the macro and thus have more control.
* I have used macros where I needed to, or where I felt it made sense. However, I've worked on this PR in drips and drabs so it's been done somewhat ad hoc. I think it's sufficient for now, but we may need to revisit this later and make things more consistent. Perhaps _everthing_ should be a macro? Or, perhaps, we should have some way (tagging in meta-data? naming convention?) to make it clear which templates are intended to be `import`ed and which `include`d.
* While I have preserved the template names, organisation and HTML output for atoms - templates, I have taken the liberty to tweak things a little for the pages. Since these are just examples, it won't affect consumers of Gravity at all.
* Going through all our templates with fresh eyes has highlighted a number of questionable choices. Some things are poorly organised (e.g. we have different kinds of lists scattered all over the place), naming is a bit confusing and the whole section organism thing is a massive clusterfuck. Once this gets merged we really should spend some time spring cleaning. :-D
* I've rejigged the JSON data a fair bit. There is now much less shared stuff in `data.json` and more per-pattern JSON data. While this does introduce some duplication, I think it makes it easier for developers to understand where data for a particular template is coming from. PL's a bit opaque in this respect unfortunately, so debugging missing or unexpected values is a royal pain in the arse.
    * A beneficial side-effect is that this side-steps some of PL's data merging quirks and makes our page examples tidier. For instance, we no longer have weird Lorem ipsum nav links appearing there!


bradarnst-buildit:
Seems a sane/good choice. I'll go look at nunjucks now.

dw-buildit:
For me, the `html` view option of pattern lab is presenting poorly formatted html and I can't see all of it in the view. e.g. http://127.0.0.1:49689/?p=templates-homepage shows a large whitespace before the first tag: 
screenshot 2019-02-11 at 10 34 16

james-nash:
Good spot @dw-buildit . I'll look into that. Nunjucks does have some whitespace control features which should let us tidy this a bit.

Things like indentation probably won't be perfect though - just as they aren't when using Mustache. So, perhaps longer term it's worth investigating if we can make a Pattern Lab plugin the beautifies the mark-up before it is output.

james-nash:
@dw-buildit I have now tweaked all the templates to eliminate redundant whitespace in the rendered HTML output. The final output is not necessarily indented nicely, but it should be no worse than it was when we were using Mustache templates.

Please review when you next get a chance and approve - or not :-P. Thanks!

dw-buildit:
Looks good now :thumbsup:

Issue #189, opened 2019-03-06
Tidy-up and clarify public API


james-nash:
We have never clearly specified and document what is (and what is not) considered this library's public API.

I believe we have a number of API surfaces:

* **JS Build-time API**: Currently, if you `require('@buildit/gravity-ui-sass');` you get a JS object with absolute file paths to key items published in the NPM package (such as our main CSS file, the SVG symbols file and meta-data). This API is strictly intended for consumers to use within their build setup - these file paths should never make their way into actual apps or websites. The idea is sound, but the API is a tad scruffy though so perhaps we can tidy it for 1.0.0
* **CSS (DOM?) API**: Our CSS requires a consuming project's DOM to structured in a specific way (certain elements, attributes and/or classes). So, in a sense, that specification is our API. A breaking change would be any change to the CSS that forces downstream projects to alter their DOM in some way.
* **SASS API**: We also distribute the SASS code. The idea is that consumers can then access the SASS vars, mixins, etc. we define in order to re-use those for any bespoke CSS code they add on top of what Gravity has out of the box.
* **Assets API** (?): Right now we provide a set of SVG symbols. If we were to remove or rename any of their IDs that could "break" downstream projects. So, this is another API surface to consider. Arguably, it's covered by the build-time API though. You can (and probably should) use that to get SVG symbol's IDs, in which case you don't need to hard-code them anywhere.

I _think_ that's it for now. So, stuff that's "private" includes the internals of our build process and anything to do with the pattern library.

It's conceivable that additional public APIs may be added in future - e.g. if we ever bundled some client-side JS.

So, what we need to do is:
* **Agree** amongst ourselves that these are indeed what we consider our public API and will monitor for breaking changes from now on.
* **Tidy-up the build-time API**. We should review what file paths we expose. Is it too much or too little? Can it be done more nicely? Also, is it correct for this to be our package's main entry point? Especially, considering that we may one day add client-side JS?
* **Document** each of these APIs (most likely in a markdown file somewhere).

PR #194, opened 2019-03-13
feat: added scss files so each itcss layer can be imported individually

danbull:
re #189

**Description**
Currently any project consuming Gravity has to include gravity scss or compiled css first and then include local, project specific css afterwards. This new way of importing scss files allows projects to consume each ITCSS layer individually and interlace project specific scss. This means the entire project can follow the ITCSS architecture.

**Checklist**
### Other


danbull:
@james-nash @dw-buildit let me know what you think, if we're happy with the structure I'll add some docs to #189

src/styleguide/_patterns/_generated/symbols.mustache
```diff
@@ -0,0 +1,11 @@
+<svg style="display: none;" aria-hidden="true" xmlns="http://www.w3.org/2000/svg">
```
james-nash: Oh. This file shouldn't be checked into git.

It gets generated by the Pattern Lab build. However, as part of the recent Nunjucks conversion, the `.gitignore` rules was also updated. This is probably a left over from builds you did while we still used Mustache.

Please remove it.

james-nash:
@danbull Btw, there's no need to update the CHANGELOG, `semantic-release` does that for us now.

danbull:
> Looks good to me. Thanks for doing this!
> 
> One comment though: Is the `_settings.all.scss` naming not a tad verbose? Why not just `_settings.scss` and so on?
> 
> That being said, I don't feel strongly about that, so approving this PR. Keep the names as they are or change them before merging - I'll leave it up to you. :-)

Yeah I had that originally but then saw Harry Roberts used `.all` and thought it might be useful - in the future we may have things like `components.forms.scss`, for example

PR #202, opened 2019-03-27
Basic forms

### LINKED ISSUE #198: Add form field components
Gravity mainly has editorial/content styling so there's not much in the way of web app type components. To get things going I'll add a form row component along with:
- label
- text/password input
- call to action button

danbull:
Closes #198

Added:
- text input
- label
- form group component
- basic example of a login form

Form group component is used to lay out an input and associated label.


src/styleguide/_patterns/01-atoms/07-forms/00-text-input.njk
```diff
@@ -0,0 +1,12 @@
+{# Reusable macro for other templates to import: #}
+{%- macro textInput(id, classNames, type, disabled) -%}
```
james-nash: Maybe provide some default values for `type` (probably `"text"`) and `disabled` (probably `false`) for convenience?

src/styleguide/_patterns/02-molecules/03-form-group/01-form-group.njk
```diff
@@ -0,0 +1,13 @@
+{%- from 'atoms-label' import label -%}
+{%- from 'atoms-text-input' import textInput -%}
+
+{# Reusable macro for other templates to import: #}
+{%- macro formGroup(form) -%}
+<div class="grav-c-form-group">
+  {{ label(form.label, form.id) }}
```
james-nash: I think it's safer to use `| safe` whenever using macros that output HTML.

While it's configurable, I think Nunjuck's default behaviour is to auto-escape strings. So, if we're not careful this might turn into `&lt;label>` rather than `<label>`.

src/ui-lib/sass/03-elements/_input.scss
```diff
@@ -0,0 +1,7 @@
+input[type=text],
+input[type=password] {
```
james-nash: Should we add other textbox style inputs here: `email`, `url`, `tel`... etc. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

In fact, it may be easier to set these defaults on any `input` and then override for the handful of exceptions like `radio` and `checkbox` (actually, they might be the only ones that don't render a text box).

src/ui-lib/sass/05-components/01-atoms/03-forms/_01-form-group.scss
```diff
@@ -0,0 +1,10 @@
+.grav-c-form-group {
+  * {
+    margin: 0;
+  }
+
+  > label {
+    display: block;
+    margin-bottom: 0.5em;
```
james-nash: Would it not be more succinct to simply override the `input`'s `margin-top`? As the first child of the group element, the `label` will already have a `margin-top: 0;`.

So, something like this:
```sass
.grav-c-form-group {
  > label {
    display: block;
  }

  > input {
    margin-top: $grav-sp-s;
  }
}
```

src/styleguide/_patterns/02-molecules/03-form-group/01-form-group.njk
```diff
@@ -0,0 +1,13 @@
+{%- from 'atoms-label' import label -%}
```
james-nash: I suggest making the folder name more generic. Perhaps `03-forms/...` rather than `03-form-group/...`. If/when we have other form related molecules we can then pop them in there too. :-)

src/styleguide/_patterns/01-atoms/07-forms/00-text-input.njk
danbull: Added default type of text. The disabled value should default to nothing (false) or "disabled" if true is passed.

src/ui-lib/sass/03-elements/_input.scss
danbull: I've gone applying styling to all `input` types and have used `:not` selectors to exclude `radio` and `checkbox`. We may want to consider adding further exceptions but I think these are the key ones for now.

PR #213, opened 2019-04-05
fix: adjust position of outline when logo is focused

### LINKED ISSUE #211: Header Logo has odd outline on focus
The header logo when, focused, has an odd outline.

I'd expect it to be a rectangle rather than this weird polygon.

See screenshot:
Screenshot 2019-04-05 at 09 41 55

Seen on FireFox 66.0.2 (64-bit).

allancorbett:
fix #211


src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -1,90 +1,92 @@
 $grav-page-heading-logo-width: 250px;
+// stylelint-disable-next-line scss/dollar-variable-pattern
+$trigger-breakpoint: gravy-breakpoint(medium);
```
james-nash: Can we rename this to `$grav-page-heading-trigger-breakpoint`? Since it's no longer within a block, it is effectively global. It don't think it clashes with anything else, but having a more specific name ought to minimise the risk of that happening in future.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -1,90 +1,92 @@
...
+    margin: $grav-sp-m $grav-sp-m 0 0;
+    flex-shrink: 0;
   }
+}
 
-  .grav-c-logo {
-    display: block;
-    max-height: 33px; // Otherwise IE11 makes it 150px tall :-(
-    fill: $grav-co-neutral-asphalt;
+.grav-c-logo {
```
james-nash: This needs to move back into the `body > header` block.

`.grav-c-logo` is actually another component, so by doing this your affecting all instances of it. However, this code is only meant to affect instances of logos that appear within a page header.

src/ui-lib/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -1,90 +1,92 @@
...
+  @media (min-width: $trigger-breakpoint) {
+    width: 100%;
   }
+}
 
-  /* stylelint-disable-next-line selector-max-compound-selectors */
-  a:hover .grav-c-logo {
-    fill: $grav-co-neutral-warm-grey;
-  }
+/* stylelint-disable-next-line selector-max-compound-selectors */
```
james-nash: Same thing here and in the following blocks. Please move this back within `body > header`, so that we don't break these components when they occur outside of a page header.

allancorbett:
> LGTM! 👍
> 
> Thanks and welcome to the hallowed ranks of Gravity contributors! :-D

😭

PR #214, opened 2019-04-05
ci(travis): enabled commitlint in Travis CI

### LINKED ISSUE #210: Travis doesn't run commit lint
**Describe the bug**
We currently only have `commitlint` configured to run as one our pre-commit hooks.

However, sometimes incorrectly formatted commit messages can still creep in - e.g. when someone has not (recently) run `npm install` and therefore doesn't have that hook setup.

To spot such erroneous commit messages sooner, we should therefore also run `commitlint` in Travis, so that PR checks fail if any of the commits messages don't follow the correct format.

It _looks_ like this was already setup at some point, but the relevant line in `.travis.yml` has been commented out: https://github.com/buildit/gravity-ui-web/blob/develop/.travis.yml#L12

@ashblue Any idea why that might be? Was it intentionally disabled?

ashblue:
Commitlint was originally turned off to prevent old commits from causing Travis CI to fail.

Closes #210

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
 install:
   - npm ci
 script:
-#  - commitlint-travis
+  - commitlint-travis
   - npm run-script styleguide
 before_deploy:
   - pip install --user awscli


allancorbett:
@ashblue I assume you want to push the magic button to merge?

james-nash:
@allancorbett Thanks. Yeah, go for it (once Travis has finished doing its thing)!

PR #218, opened 2019-04-08
fix(page-header): fixes linter issues

james-nash:
**Description**
A few whitespace issues that the SASS linter doesn't like snuck in. This fixes those.


allancorbett:
@james-nash merged and deleted the branch

Issue #147, opened 2018-06-20
Implement more systematic use of colour


james-nash:
Our modest 4 page website designs already include many different styles of links and also a button. The use of colour for those interactive elements is very ad hoc as we never defined any clear guidance or rules.

This story is to establish and implement a more systematic use of colour with the goal of encouraging (enforcing?) more consistent usage in the future.

james-nash:
@buildit/gravity-maintainers

After much thought and experimentation, I'm now leaning towards a solution along these lines:

## Concept
* Define a smallish set of CSS custom properties representing colour _purposes_
    * There should also be some rules on which pairs of colours are allowed to be combined into foreground / background pairs.
* Define one or more colour _themes_, which are assignments of colour _values_ (taken from our brand palette) to those custom properties.
    * The colour assignments should ensure that each permitted pairing has a contrast ratio that meets at least WCGA 2.0 AA criteria
    * There is at least one, "default", colour theme which is essentially set on via the `body` (or perhaps `:root`) selector in CSS.
    * Every other theme would have a class selector, which simply overrides the _values_ of the custom properties
* Downstream, in the styles for UI components - colours must always reference those CSS properties (or, alternatively, use `inherit` or `currentColor`)

Using this approach, _every_ UI component would become themable. Likewise, it would become very easy to add more themes (or to temporarily alter our UI colours - e.g. for seasonal events).

## Prototype
I have put together a prototype of this in CodePen here: https://codepen.io/cirrus/pen/wxVqxG

At the time of writing, this prototype has 10 colour purposes:
```css
  --grav-co-fg
  --grav-co-bg

  --grav-co-fg-secondary
  --grav-co-bg-secondary
  
  --grav-co-fg-primary
  --grav-co-bg-primary
  
  --grav-co-fg-tertiary
  --grav-co-bg-tertiary
  
  --grav-co-fg-neutral
  --grav-co-bg-neutral
```

`--grav-co-fg` is the default foreground colour. It is allowed to be paired up with any of the 5 `--grav-co-bg*` colours.

Similarly, `--grav-co-bg` is the default background colour and may be paired with any of the 5 `--grav-co-fg*` colours.

Then, each theme _should_ assign colours to each of those 10 custom properties such that all permitted pairings meet AA standards.

The "primary", "secondary", "tertiary" and "neutral" colours are intended to be applied in the UI as follows:

[...]

PR #220, opened 2019-04-10
feat: dark mode

allancorbett:
BREAKING CHANGE: it could all go horribly wrong...

See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme for more info. 

Safari and FF67 support it at the moment with Chrome going to support it. 

Screenshot 2019-04-10 at 10 15 42

Screenshot 2019-04-10 at 10 15 56

**Things still needing done:**
- [x] "What we do" images
- [ ] General discussion about if we want to support this or not...

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,12 +43,16 @@
     "url": "http://buildit.wiprodigital.com/"
   },
   "contributors": [
-    "Matteo Pescarin (http://peach.smartart.it)",
+    "Gregg Anderson",
+    "Charles \"Ash Blue\" Brindle",
+    "Daniel Bull",
+    "Allan Corbett (http://superallan.com)",
+    "Valerio Francescangeli",
+    "James Nash (http://cirrus.twiddles.com/)",
     "Ben Oakenfull",
     "Marcos Peebles",
-    "James Nash (http://cirrus.twiddles.com/)",
-    "Valerio Francescangeli",
-    "Daniel Bull"
+    "Matteo Pescarin (http://peach.smartart.it)",
+    "Daniel Wilkinson"
   ],
   "license": "MIT",
   "bugs": {
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "semantic-release": "^15.13.3",
     "styleguidekit-assets-default": "^3.5.2",
     "styleguidekit-mustache-default": "^3.1.0",
-    "stylelint": "^9.9.0",
+    "stylelint": "^10.0.0",
     "stylelint-config-sass-guidelines": "^5.3.0",
     "stylelint-order": "^2.1.0",
     "stylelint-scss": "^3.4.4"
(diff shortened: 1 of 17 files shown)


james-nash:
Thanks for proposing & implementing this feature!

I think it's a good addition. It's something we were thinking of possibly doing in `next` via the new color system (see #147), but we never really thought about doing it for the _current_ 1.x version. However, why not? Especially since you've pretty much done it already.

So, I propose:
1. Polish this off and let's get it released
1. We'll update the current Buildit website to use this (and figure out what to do about those "What we do" images* in the process
1. In `next`, once the new color system goes in, we'll retrofit an equivalent dark mode (should be pretty easy to implement)

@danbull @dw-buildit : What say ye?


*) Off the top of my head we could:
* Inline those SVGs and set their fill to `currentColor` so they automagically adapt
* Produce white versions of them and use a `<picture>` element to select the light or dark image based on the `prefers-color-scheme` media query.

allancorbett:
> Inline those SVGs and set their fill to currentColor so they automagically adapt

That would be my preference.

allancorbett:
So I've implented the dark mode for certain images using some (probably rather ropey) methods: `src/ui-lib/sass/05-components/01-atoms/04-media/_24-image.scss`

Thoughts?

package.json
```diff
@@ -43,16 +43,12 @@
     "url": "http://buildit.wiprodigital.com/"
   },
   "contributors": [
-    "Gregg Anderson",
```
james-nash: Not sure why this is showing up as a change versus `develop`. We need to be careful when this PR eventually gets merged that it doesn't revert the recent update to the contributors list.

src/ui-lib/sass/05-components/01-atoms/04-media/_24-image.scss
```diff
@@ -0,0 +1,7 @@
+.grav-c-image {
+  &__invertible {
+    @media (prefers-color-scheme: dark) {
+      filter: invert(100);
```
james-nash: I think this is fine. Especially since it's opt-in via the modifier class.

james-nash:
Looks good to me - just watch out with the `package.json` because it looks like it might overwrite the recent updates to that.

package.json
allancorbett: Not sure - fixed in another commit though.

src/ui-lib/sass/05-components/01-atoms/04-media/_24-image.scss
allancorbett: 👍

james-nash:
This is lookin' good to me.

I'm happy for this PR to be un-drafted (is that the right term?) and merged.

I see you've already prepped some associated changes on the website, so let's get this released as soon as we can. :-)

PR #221, opened 2019-04-12
feat(deployment): add cdn publishing

### LINKED ISSUE #191: Publish UI library copy to CDN
Feature request: In addition to distributing Gravity's CSS library via NPM, it would be handy to also publish to a CDN. This would allow consumers to simply add something like `<link rel="stylesheet" href="https://static.buildit.digital/gravity/v1.0.0/gravity.css">` to an HTML file to get started instead of needing to faff about with NPM installs. (and, if we add JS later it could be similar)

I imagine this could be useful for small projects and prototypes since it reduces the intial time and effort to get up and running.

For public sites that only need the out-of-the-box Gravity styles & components, there may even be a performance benefit of referencing a central CDN copy instead of hosting their own versions of those files. As users visit various Buildit properties, those files will have already been cached, so visits to new pages referencing them should load a little faster.

Note that this would be _in addition_ to the current NPM distribution, so more complex use-cases can still do what they do today.

I'd imagine this would be an additional deployment step in our CI builds that uploads a copy of the pre-compiled CSS, JS and assets to a web server. This should coincide with NPM publishes - i.e. whenever we issue a new release.

This would be an _additive_ process - i.e. previously released versions remain there. Therefore, each new release should be published to a unique URL. Using the version number in the URL's path seems like an obvious choice. Something along these lines perhaps:

* `https://static.buildit.digital/gravity/v1.0.0/...` (where `v1.0.0` corresponds to that release's version)

That way consumers can link to a specific version. Also, the URL is human-readable, which makes it easy to understand what your linking to.

We could configure the HTTP response headers to have a very long cache time (or even mark those resources as "immutable"), so that proxies and browsers are encouraged to not re-request those files again - after all, once published, a release will never change.

Some potential nice-to-haves might include:

* Ability to link to the latest release of a major version. For example, something like: `https://static.buildit.digital/gravity/v1.x/...` could always return whatever the most recently release `v1.x.y` release is. This would allow consumers to always receive the most recent _compatible_ release.
     * For these URLs we'd need to be careful with the cache-related response headers. Clients and proxies should at least revalidate these URLs after a few times so that they successfully pick up new releases.
* Index page(s). It might be a nice touch if visiting `https://static.buildit.digital/gravity/` would show a page containing a list of all currently published releases.

thehammeranderson:
Closes #191

**Description**
Add publishing of master branch *.css, *.js, and an index.html to an S3 backed CloudFront CDN that
stores versions (i.e. https://static.buildit.digital/gravity/v1.0.1/gravity.css)

Will demo before merge

### DIFF
--- a/cdn-index.template
+++ b/cdn-index.template
@@ -1,7 +1,7 @@
 <html>
 <head><link rel="stylesheet" href="https://static.buildit.digital/gravity/v1.0.1/gravity.css"></head>
 <body>
-<h2>Versions</h2>
+<h1>Versions</h1>
 <ul>{versionlist}</ul>
 <a href="https://github.com/buildit/gravity-ui-web/blob/develop/CHANGELOG.md">See change log</a>
 </body>
--- a/cdn-index.template
+++ b/cdn-index.template
@@ -1,0 +1,8 @@
+<html>
+<head><link rel="stylesheet" href="https://static.buildit.digital/gravity/v1.0.1/gravity.css"></head>
+<body>
+<h2>Versions</h2>
+<ul>{versionlist}</ul>
+<a href="https://github.com/buildit/gravity-ui-web/blob/develop/CHANGELOG.md">See change log</a>
+</body>
+</html>
--- a/scripts/cdn_publish.sh
+++ b/scripts/cdn_publish.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
+cdnGravityDir="gravity-ui-web"
+cdnDir="cdn-dist"
 version="$(git describe --abbrev=0 --tags)"
-mkdir -p cdn-dist/gravity/${version}
+mkdir -p ${cdnDir}/${cdnGravityDir}/${version}
 
-cp dist/ui-lib/* cdn-dist/gravity/${version}
+cp dist/ui-lib/* ${cdnDir}/${cdnGravityDir}/${version}
 
 versionList=""
 for version in $(git tag -l --sort=-committerdate); do
@@ -10,7 +12,7 @@ for version in $(git tag -l --sort=-committerdate); do
 done
 
 file_contents=$(<./cdn-index.template)
-echo "${file_contents//\{versionlist\}/$versionList}" > cdn-dist/gravity/index.html
+echo "${file_contents//\{versionlist\}/$versionList}" > ${cdnDir}/${cdnGravityDir}/index.html
 
-~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8" --cache-control "max-age=0"
-~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*.html" --exclude ".DS_Store" --cache-control "max-age=3153600000, immutable"
+aws s3 sync ${cdnDir} s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8" --cache-control "max-age=0"
+aws s3 sync ${cdnDir} s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*.html" --exclude ".DS_Store" --cache-control "max-age=3153600000, immutable"
--- a/scripts/cdn_publish.sh
+++ b/scripts/cdn_publish.sh
@@ -13,4 +13,4 @@ file_contents=$(<./cdn-index.template)
 echo "${file_contents//\{versionlist\}/$versionList}" > cdn-dist/gravity/index.html
 
 ~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8" --cache-control "max-age=0"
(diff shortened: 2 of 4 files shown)


james-nash:
Sweet!

Thanks so much for setting this up @thehammeranderson. I look forward to the demo.

cdn-index.template
```diff
@@ -0,0 +1,8 @@
+<html>
+<head><link rel="stylesheet" href="https://static.buildit.digital/gravity/v1.0.1/gravity.css"></head>
+<body>
+<h2>Versions</h2>
```
james-nash: Could you make this an `h1` please? It's the top level heading of this page, so we should use the appropriate semantics. :-)

scripts/cdn_publish.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+version="$(git describe --abbrev=0 --tags)"
+mkdir -p cdn-dist/gravity/${version}
+
+cp dist/ui-lib/*.css cdn-dist/gravity/${version}
```
james-nash: You can just copy everything from this folder (i.e. `dist/ui-lib/*`). Things like the SVG (and it's associated JSON metadata file) should also be available to consumers.

In future Gravity releases we might also begin to add other files (e.g. fonts).

scripts/cdn_publish.sh
```diff
@@ -0,0 +1,18 @@
...
+
+versionList=""
+for version in $(git tag -l --sort=-committerdate); do
+  versionList+="<li>${version}</li>"
+done
+
+file_contents=$(<./cdn-index.template)
+echo "${file_contents//\{versionlist\}/$versionList}" > cdn-dist/gravity/index.html
+
+~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --delete --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8"
```
james-nash: Won't `--delete` remove any previously published files? If so, I'm don't think that's what we want.

I had imagined this to be cumulative, each new Gravity release would be _added_ here, but all previously published one would remain available.

scripts/cdn_publish.sh
```diff
@@ -0,0 +1,18 @@
...
+~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --delete --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.css" --content-type "text/css; charset=utf-8"
+~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --delete --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.js" --content-type "application/javascript; charset=utf-8"
```
james-nash: Perhaps we can tweak the `cache-control` parameters for each Gravity release's files (i.e. everything under `/gravity/${version}/`)?

Based on this article: https://csswizardry.com/2019/03/cache-control-for-civilians/ , I wonder if `immutable` could be appropriate here (signalling to browsers & proxies that they never need to revalidate those resources).

If we get it right, it could provide a small performance boost to any projects referencing these CDN files.

scripts/cdn_publish.sh
thehammeranderson: since we are creating version sub directories dynamically, you are probably right.  I'll remove that flag.  good catch

scripts/cdn_publish.sh
thehammeranderson: Would we want to keep them separate to allow for us to define content-type?  Not sure how important that is for us

thehammeranderson:
all requested changes have been implemented

scripts/cdn_publish.sh
```diff
@@ -0,0 +1,16 @@
...
+versionList=""
+for version in $(git tag -l --sort=-committerdate); do
+  versionList+="<li>${version}</li>"
+done
+
+file_contents=$(<./cdn-index.template)
+echo "${file_contents//\{versionlist\}/$versionList}" > cdn-dist/gravity/index.html
+
+~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8" --cache-control "max-age=0"
+~/.local/bin/aws s3 sync cdn-dist s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*.html" --cache-control "max-age=3153600000, immutable"
```
james-nash: Hmm... this kinda works - the CSS & JS files are being served up with the correct mime-types, even though we don't explicitly set them here. However, they don't have any character encoding set. Is it possible to set that to UTF-8 - at least for text-based types like CSS, JS, JSON and SVG?

While it shouldn't make much practical difference right now, we might get some weirdness if we were to add non-ASCII chars to any of them (e.g. emoji).

thehammeranderson: Yes, it will assume file type based on extension.  If you want encoding, we need to separate them out.

scripts/cdn_publish.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+version="$(git describe --abbrev=0 --tags)"
```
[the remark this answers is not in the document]
thehammeranderson: yep

thehammeranderson:
> A couple of minor tweaks (sorry!).
> 
> Also, a query for @ashblue. If we merge this before his one, we don't necessarily need to change anything right now though.

We will either need to wait for his PR to merge, or go back and fix one line in the cdn script to copy from the new directory after the merge.

PR #223, opened 2019-04-16
Release/dark side

james-nash:
Releases some recently merged PRs including *dark mode*!

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -12,7 +12,7 @@
         "ignore": [ "attribute" ]
       }
     ],
-    "max-nesting-depth": 2,
+    "max-nesting-depth": 3,
     "color-named": [
       "never",
        {
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,18 +14,18 @@
       }
     },
     "@babel/core": {
-      "version": "7.3.4",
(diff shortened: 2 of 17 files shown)


.stylelintrc
```diff
@@ -12,7 +12,7 @@
         "ignore": [ "attribute" ]
       }
     ],
-    "max-nesting-depth": 2,
+    "max-nesting-depth": 3,
```
dw-buildit: where do we get our stylelint config from?

i'm not against this change, just wondered what our process and/or overall guidelines are

james-nash: They're home grown, and we don't really have a process in place for when/how to update them. We probably should :-)

FWIW, we have some coding conventions documented here: https://github.com/buildit/gravity-ui-web/blob/develop/docs/naming-conventions.md  ...and some of those (e.g. SASS variable names) are enforced by our custom stylelint rules.

I think that this change is fine for now.

However, at some point (soon?), it might be good to review our coding conventions and update/amend rules as we see fit. Relatedly, once the Lerna stuff goes in, I'd like us to split out the stylelint config into an NPM package of its own, so that it can be used by other projects too.

allancorbett: That specific change was because it was ludicrously difficult to target an element. 3 deep nesting seems not too extravagant imo.

PR #224, opened 2019-04-16
Feature/lerna wrapper

ashblue:
Here it is. The Lerna wrapper to make this a Lerna friendly repo for multiple packages. Please take a good hard look at the paths below. Note this only converts the core package, but breaking up gravity-ui into a new package or starting a new one should be pretty simple.

* Root: Mostly a simple Lerna wrapper around the whole project. I strongly recommend reviewing the travis.yml in detail. Some slight modifications to account for dealing with a nested repo for deployment.
* `packages/semantic-release/*`: All of the packages here should be both sustainable and reusable. Important for moving forward as it includes custom Lerna modifications to the normal Semantic Release workflow.

NOTE: Most changes listed are just moved files.


README.md
```diff
@@ -1,237 +1,14 @@
-# gravity-ui-web
+# Gravity UI
 
-[![Greenkeeper badge](https://badges.greenkeeper.io/buildit/gravity-ui-web.svg)](https://greenkeeper.io/)
+This project is a [Lerna](https://lernajs.io/) monorepo composed of several sub-packages. Everything is binded
+together with a custom [Semantic Release](https://github.com/semantic-release/semantic-release) pipeline configuration.
 
-Library of styles, components and associated assets to build UIs for the web. Part of Buildit's Gravity design system.
+[![Commitizen friendly](https://img.shields.io/badge/commitizen-friendly-brightgreen.svg)](http://commitizen.github.io/cz-cli/)
```
dw-buildit: I think it would be good to link to the packages that are within this monorepo. As this readme is likely to be the landing point still for most people and then they can click through to the correct sub-package (and readme) from here

packages/gravity-ui/package.json
```diff
@@ -0,0 +1,99 @@
...
+    "name": "Buildit",
+    "url": "http://buildit.wiprodigital.com/"
+  },
+  "contributors": [
+    "Matteo Pescarin (http://peach.smartart.it)",
+    "Ben Oakenfull",
+    "Marcos Peebles",
+    "James Nash (http://cirrus.twiddles.com/)",
+    "Valerio Francescangeli",
+    "Daniel Bull"
```
dw-buildit: We'll have to be a little careful about this merge as some of the gravity-ui-web files will have changed (such as this package.json) and I don't think this will get flagged up as a conflict (or it was, and it was resolved in the base package.json file).

https://github.com/buildit/gravity-ui-web/blob/develop/package.json

packages/semantic-release/config/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@buildit/semantic-release",
+  "version": "1.0.0",
+  "description": "> TODO: description",
+  "author": "Ash Blue <ash@blueashes.com>",
+  "homepage": "https://github.com/buildit/gravity-ui-sass#readme",
```
dw-buildit: Minor: this and the description should probably be updated ;)

packages/semantic-release/lerna-analyzer/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@buildit/semantic-release-lerna-analyzer",
+  "version": "1.0.0",
+  "description": "> TODO: description",
+  "author": "Ash Blue <ash@blueashes.com>",
+  "homepage": "https://github.com/buildit/gravity-ui-sass#readme",
```
dw-buildit: Minor: homepage and description should be updated

packages/semantic-release/release-notes-generator/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@buildit/semantic-release-lerna-release-notes-generator",
+  "version": "1.0.0",
+  "description": "> TODO: description",
+  "author": "Ash Blue <ash@blueashes.com>",
+  "homepage": "https://github.com/buildit/gravity-ui-sass#readme",
```
dw-buildit: Minor: homepage and description should be updated

README.md
james-nash: I agree. Just a bullet point list of links would suffice for now I think.

packages/gravity-ui/.gitignore
```diff
@@ -0,0 +1,122 @@
+# Exclude build Output
```
james-nash: Would it perhaps be better to use either the full `gravity-ui-web` name for this directory? Feels more consistent.

Also, it's conceivable that other `gravity-ui-*` packages may one day live in this monorepo. `gravity-ui-nunjucks` and/or `gravity-ui-react` perhaps?

packages/semantic-release/config/package.json
james-nash: ...and the homepage. Ought to be a link to this package's README as opposed to the monorepo's root README.

packages/semantic-release/lerna-analyzer/.gitignore
```diff
@@ -0,0 +1 @@
+dist/**/*
```
james-nash: I think just `dist/` is sufficient

README.md
ashblue: Added a list of currently maintained packages

packages/gravity-ui/package.json
ashblue: It should be tracking the nested package.json to the one currently in `develop`'s root. I may have mis-merged when there was a conflict due to a large number of changes in the dependencies when I started tearing out pieces to go in the root (good catch). I've moved over the missing names.

packages/gravity-ui/.gitignore
ashblue: Changed to `gravity-ui-web`

.travis.yml
```diff
@@ -17,24 +18,24 @@ deploy:
...
-      ~/.local/bin/aws s3 sync dist s3://${STAGING_BUCKET} --delete --region=${STAGING_BUCKET_REGION} --exclude "*" --include "*.js" --content-type "application/javascript; charset=utf-8" &&
-      ~/.local/bin/aws s3 sync dist s3://${STAGING_BUCKET} --delete --region=${STAGING_BUCKET_REGION} --include "*" --exclude "*.js" --exclude "*.html" --exclude "*.css"
+      ~/.local/bin/aws s3 sync dist s3://${STAGING_BUCKET} --delete --region=${STAGING_BUCKET_REGION} --exclude "*" --include "packages/gravity-ui-web/*.html" --content-type "text/html; charset=utf-8" &&
+      ~/.local/bin/aws s3 sync dist s3://${STAGING_BUCKET} --delete --region=${STAGING_BUCKET_REGION} --exclude "*" --include "packages/gravity-ui-web/*.css" --content-type "text/css; charset=utf-8" &&
+      ~/.local/bin/aws s3 sync dist s3://${STAGING_BUCKET} --delete --region=${STAGING_BUCKET_REGION} --exclude "*" --include "packages/gravity-ui-web/*.js" --content-type "application/javascript; charset=utf-8" &&
+      ~/.local/bin/aws s3 sync dist s3://${STAGING_BUCKET} --delete --region=${STAGING_BUCKET_REGION} --include "*" --exclude "packages/gravity-ui-web/*.js" --exclude "*.html" --exclude "*.css"
```
james-nash: Hmm... upon closer inspection, are these right?

Is it not `packages/gravity-ui-web/dist` that needs to be synced now, rather than just `dist`? I presume then don't need the path later on for the `--include` and `--exclude` bits (I _assume_ those are relative to the dir being synced).

So, putting it together, something like this:
```
 ~/.local/bin/aws s3 sync packages/gravity-ui-web/dist s3://${PROD_BUCKET} --delete --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8" &&
...
```

Apologies for not spotting this one sooner.

README.md
```diff
@@ -1,237 +1,25 @@
-# gravity-ui-web
+# Gravity UI
 
-[![Greenkeeper badge](https://badges.greenkeeper.io/buildit/gravity-ui-web.svg)](https://greenkeeper.io/)
```
james-nash: Can we keep the Greenkeeper badge please?

dw-buildit: Is greenkeeper working on the overall monorepo? The badge is still in the README.md for gravity-ui-web... https://github.com/buildit/gravity-ui-web/blob/feature/lerna-wrapper/packages/gravity-ui-web/README.md

james-nash: I _think_ so.

A quick glance at their docs confirms that they do support monorepos. But it looks like we might need to add a `greenkeeper.json` config file for it to work properly: https://greenkeeper.io/docs.html#monorepo

I'd suggest we gets this merged and see what happens. If Greenkeeper stops working, we can always add that config file later.

.travis.yml
ashblue: Good catch here, looks like I edited the wrong path with the AWS commands.

README.md
ashblue: I've added back the back to the root `README.md`. I'm suspicious Greenkeeper is only going to run against the root `package.json`, but hopefully I'm wrong.

james-nash:
@thehammeranderson If you get a chance, could you check the most recent commit please?

I've (hopefully) updated `cdn_publish.sh` to correctly work with prefixed tags. So far all our tags were `v1.2.3` form. But, once this PR gets released, they will be `[package-name]-v1.2.3`.

_However_, we only want to publish stuff from the `gravity-ui-web` package to CDN, so we need to not only match prefixed tag names, but also avoid matching ones belonging to other packages.

I've "tested" this locally by adding some tags like `gravity-ui-web-v1.2.3` and `other-package-v1.2.3` and it seems to correctly omit `other-package-v1.2.3` :-)

Nonetheless, a second pair of eyes would be appreciated. Thanks!

scripts/cdn_publish.sh
```diff
@@ -1,16 +1,19 @@
 #!/bin/bash
 cdnGravityDir="gravity-ui-web"
 cdnDir="cdn-dist"
-version="$(git describe --abbrev=0 --tags --match *v[0-9]*.[0-9]*.[0-9]*)"
+allVersions=`git tag --list --sort=-committerdate | grep -E '^(gravity-ui-web-)?v[0-9]*.[0-9]*.[0-9]*'`
```
dw-buildit: @james-nash Should there be something in that cdn code to bail out if it’s _not_ `gravity-ui-web` being committed?

If i updated a version of something else, would that code still run every time? Publishing a non-updated `gravity-ui-web` cdn folder?

james-nash: Good point!

Any chance you could have a creack at it? I'm busy with the new color system PR at the moment.

dw-buildit: Had a look and a think and as there's extra considerations we can leave this as identified technical debt

Issue #235, opened 2019-04-26
Make CDN publish script bail out for non-CDN published packages


james-nash:
With the scwitch to a Lerna-based monorepo in PR #224, this repo contains the sources for multiple NPM packages.

Currently, as part of _any_ release, the `scripts/cdn_publish.sh` script will be run. Its purpose is to publish assets from the `@buildit/gravity-ui-web` package to a CDN.

It already ignores versions of other packages when it composes its `index.html` page. However, it will still make that page and upload it every time.

However, if it is only other package(s) being published, it's redundant for it to do those steps. See review comments here: https://github.com/buildit/gravity-ui-web/pull/224#pullrequestreview-231055578

**Describe the solution you'd like**
The script should bail out if only other packages are being published.

dw-buildit:
When we push to `master` (where this script is triggered in travis.yml) there could, in theory, be multiple tagged versions of multiple sub-packages. This needs to work within that framework.

I wonder if there's a more elegant solution on using tags to determine travis scripts as well as `master`. I haven't checked travis' docs but there might be a way of adding more conditions to each script block and/or firing multiple script blocks if multiple conditions are met.

PR #242, opened 2019-04-29
chore(cdn_publish.sh): only publish to CDN for gravit-ui-web changes

thehammeranderson:
ISSUES CLOSED: #235

**Description**
Only publish to CDN if gravity-ui-web source changes are committed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,6 @@ typings/
 
 # DynamoDB Local files
 .dynamodb/
+
+# Mac system files
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -85,6 +85,3 @@ typings/
 
 # DynamoDB Local files
 .dynamodb/
-
-# Mac system files
-.DS_Store
--- a/scripts/cdn_publish.sh
+++ b/scripts/cdn_publish.sh
@@ -1,26 +1,38 @@
 #!/bin/bash
 cdnGravityDir="gravity-ui-web"
 cdnDir="cdn-dist"
-allVersions=`git tag --list --sort=-committerdate | grep -E '^(gravity-ui-web-)?v[0-9]*.[0-9]*.[0-9]*'`
 
-# First tag in the list will be the latest one, but may be prefixed (e.g. "gravity-ui-web-v1.2.3")
-version=`echo $allVersions | cut -d" " -f1`
-# be sure we get only the version part and discard any prefix
-version=`expr $version : '.*\(v[0-9].*\)'`
+if [ ! -z "$TRAVIS_COMMIT_RANGE"]
+then
+  # TRAVIS_COMMIT_RANGE has a bug that includes commits outside the PR, so fix with known workaround that is future safe
+  cdnChanges=`git diff --pretty="format:" --name-only ${TRAVIS_COMMIT_RANGE/.../..} | grep -E '^packages/gravity-ui-web/src*'`
+fi
 
-mkdir -p ${cdnDir}/${cdnGravityDir}/${version}
+if [ ! -z "$cdnChanges" ]
+then
+  allVersions=`git tag --list --sort=-committerdate | grep  -E '^(gravity-ui-web-)?v[0-9]*.[0-9]*.[0-9]*'`
 
-cp packages/gravity-ui-web/dist/ui-lib/* ${cdnDir}/${cdnGravityDir}/${version}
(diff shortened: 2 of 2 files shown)


scripts/cdn_publish.sh
```diff
@@ -1,26 +1,38 @@
...
-allVersions=`git tag --list --sort=-committerdate | grep -E '^(gravity-ui-web-)?v[0-9]*.[0-9]*.[0-9]*'`
 
-# First tag in the list will be the latest one, but may be prefixed (e.g. "gravity-ui-web-v1.2.3")
-version=`echo $allVersions | cut -d" " -f1`
-# be sure we get only the version part and discard any prefix
-version=`expr $version : '.*\(v[0-9].*\)'`
+if [ ! -z "$TRAVIS_COMMIT_RANGE"]
+then
+  # TRAVIS_COMMIT_RANGE has a bug that includes commits outside the PR, so fix with known workaround that is future safe
+  cdnChanges=`git diff --pretty="format:" --name-only ${TRAVIS_COMMIT_RANGE/.../..} | grep -E '^packages/gravity-ui-web/src*'`
```
james-nash: Perhaps `packages/gravity-ui-web/*` is more appropriate. There files in package's root dir that we might change and release - not just stuff under `src/`

scripts/cdn_publish.sh
```diff
@@ -1,26 +1,38 @@
 #!/bin/bash
 cdnGravityDir="gravity-ui-web"
 cdnDir="cdn-dist"
-allVersions=`git tag --list --sort=-committerdate | grep -E '^(gravity-ui-web-)?v[0-9]*.[0-9]*.[0-9]*'`
 
-# First tag in the list will be the latest one, but may be prefixed (e.g. "gravity-ui-web-v1.2.3")
-version=`echo $allVersions | cut -d" " -f1`
-# be sure we get only the version part and discard any prefix
-version=`expr $version : '.*\(v[0-9].*\)'`
+if [ ! -z "$TRAVIS_COMMIT_RANGE"]
```
dw-buildit: I believe there should be a space before the closing test brace `if [ ! -z "$TRAVIS_COMMIT_RANGE" ]`

scripts/cdn_publish.sh
dw-buildit: Also according to this: https://github.com/travis-ci/travis-ci/issues/4596 the bug is fixed and merged. So we could probably just use `$TRAVIS_COMMIT_RANGE` instead of `${TRAVIS_COMMIT_RANGE/.../..}`

dw-buildit: Should the grep (`grep -E '^packages/gravity-ui-web/src*'`) drop the `src` on the end? Otherwise changes to package.json won't cause a publish for example.

dw-buildit: So the whole line might be:

``cdnChanges=`git diff --pretty="format:" --name-only $TRAVIS_COMMIT_RANGE | grep -E '^packages/gravity-ui-web/*'` ``

dw-buildit: Oh, snap.

.gitignore
```diff
@@ -85,3 +85,6 @@ typings/
 
 # DynamoDB Local files
 .dynamodb/
+
+# Mac system files
+.DS_Store
```
ashblue: This should probably be ignored in your Git global ignore instead of checking into the repo. https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

scripts/cdn_publish.sh
thehammeranderson: ok

scripts/cdn_publish.sh
thehammeranderson: I see the issue closed but no merges to fix the issue.  Folks continue to workaround

scripts/cdn_publish.sh
thehammeranderson: I don't see any bug fix, only a closed issue.  Folks appear to continue to work around it.

thehammeranderson: Looks like the Stale Bot closed it without resolution

dw-buildit: You're right. I was looking at this with (another) Dan and we saw some merged changes which did exactly this but looking again they're in a different package so it's not a fix - it's a work around that someone else is using 🤦‍♂

Issue #237, opened 2019-04-29
Add latest version URL to CDN


james-nash:
**Is your feature request related to a problem? Please describe.**
A copy of Gravity's releases is published to a CDN here: https://static.buildit.digital/gravity-ui-web/

Each version gets a unique URL, e.g. https://static.buildit.digital/gravity-ui-web/v2.2.0/gravity.css

It would be nice if we _additionally_ had URLs that:

* Always point to the latest version, e.g.: `https://static.buildit.digital/gravity-ui-web/latest/...`
* Alaways point to the latest, SemVer-compatible version for each major version number. E.g.:
    * `https://static.buildit.digital/gravity-ui-web/v1.x/...`
    * `https://static.buildit.digital/gravity-ui-web/v2.x/...`
    * etc.

The use-case would be for simple websites or prototypes that want to use the CDN but don't necessarily want to be pegged to a specific release.

**Describe the solution you'd like**
Add URL as per the description. However, one question would be the exact behaviour these URLs should have:

* Should they be temporary (302) HTTP redirect to whatever the latest (compatible) release's URL is? (Note that permanent (301) redirects are inappropriate in this case, since the final location will change over time, so we don't want browsers & search engines storing the final URL)
* Should they return the content directly (so be 200 responses), but set caching headers in such a way that user-agents will revalidate after a sensible period (a few days perhaps - we don't issue new releases _that_ frequently after all)

thehammeranderson:
These asks may be in conflict.  The cache header is attached to the file when it is uploaded, so you get one chance to set it.  If you are going to redirect from a url to that file, you will get that immutable cache expiration and latest really won't mean latests the next time it updates.  The only way I see latest working is by having a physical location with its own cache settings.

PR #246, opened 2019-05-02
chore(cdn_publish.sh): add latest and latest major dirs

thehammeranderson:
ISSUES CLOSED: #237

**Description**
Adds latest directory in CDN to point to to always get the latest version of artifacts as well as latest major version (1.x, 2.x, etc) directories

### DIFF
--- a/scripts/cdn_publish.sh
+++ b/scripts/cdn_publish.sh
@@ -39,8 +39,8 @@ then
 
   majorVersion=`echo $version | cut -dv -f2 | cut -d. -f1`
 
-  mkdir -p ${cdnLatestVersionsGravityDir}/${majorVersion}.x
-  cp packages/gravity-ui-web/dist/ui-lib/* ${cdnLatestVersionsGravityDir}/${majorVersion}.x
+  mkdir -p ${cdnLatestVersionsGravityDir}/v${majorVersion}.x
+  cp packages/gravity-ui-web/dist/ui-lib/* ${cdnLatestVersionsGravityDir}/v${majorVersion}.x
   mkdir -p ${cdnLatestVersionsGravityDir}/latest
   cp packages/gravity-ui-web/dist/ui-lib/* ${cdnLatestVersionsGravityDir}/latest
 
--- a/scripts/cdn_publish.sh
+++ b/scripts/cdn_publish.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 cdnGravityDir="gravity-ui-web"
-cdnDir="cdn-dist"
+cdnDistDir="cdn-dist"
+cdnCurrentVersionDir="$cdnDistDir/currentVersion"
+cdnCurrentVersionGravityDir="$cdnCurrentVersionDir/$cdnGravityDir"
+cdnLatestVersionsDir="$cdnDistDir/latestsVersions"
+cdnLatestVersionsGravityDir="$cdnLatestVersionsDir/$cdnGravityDir"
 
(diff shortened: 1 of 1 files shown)


scripts/cdn_publish.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 cdnGravityDir="gravity-ui-web"
-cdnDir="cdn-dist"
+cdnDistDir="cdn-dist"
+cdnCurrentVersionDir="$cdnDistDir/currentVersion"
+cdnCurrentVersionGravityDir="$cdnCurrentVersionDir/$cdnGravityDir"
+cdnLatestVersionsDir="$cdnDistDir/latestsVersions"
```
dw-buildit: Minor typo: "latest**s**"

Also minor request, I think file/folder names in the gravity repo tend to be hyphen cased rather than camel cased. Might be nice to keep to that even though it would only be a cosmetic difference in this case: `current-version` and `latest-version`

scripts/cdn_publish.sh
```diff
@@ -16,23 +20,34 @@ then
...
+  
+  mkdir -p ${cdnCurrentVersionGravityDir}/${version}
+  cp packages/gravity-ui-web/dist/ui-lib/* ${cdnCurrentVersionGravityDir}/${version}
 
   versionList=""
   for thisVersion in $allVersions; do
     versionList+="<li>`expr $thisVersion : '.*\(v[0-9].*\)'`</li>"
   done
-
+
```
dw-buildit: `trailing white space` 🙀

scripts/cdn_publish.sh
thehammeranderson: Doh!

scripts/cdn_publish.sh
```diff
@@ -16,23 +20,34 @@ then
...
+  echo "${file_contents//\{versionlist\}/$versionList}" > ${cdnCurrentVersionGravityDir}/index.html
+
+  ~/.local/bin/aws s3 sync ${cdnCurrentVersionDir} s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.html" --content-type "text/html; charset=utf-8" --cache-control "max-age=0"
+  ~/.local/bin/aws s3 sync ${cdnCurrentVersionDir} s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.css" --content-type "text/css; charset=utf-8" --cache-control "max-age=3153600000, immutable" 
+  ~/.local/bin/aws s3 sync ${cdnCurrentVersionDir} s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --exclude "*" --include "*.js" --content-type "application/javascript; charset=utf-8" --cache-control "max-age=3153600000, immutable"
+  ~/.local/bin/aws s3 sync ${cdnCurrentVersionDir} s3://${CDN_BUCKET} --region=${PROD_BUCKET_REGION} --include "*" --exclude "*.js" --exclude "*.html" --exclude "*.css" --cache-control "max-age=3153600000, immutable" 
+
+  majorVersion=`echo $version | cut -dv -f2 | cut -d. -f1`
+
+  mkdir -p ${cdnLatestVersionsGravityDir}/${majorVersion}.x
```
james-nash: Can we change the path here and n the next line to add the `v` prefix back. I think it'd be neater if each latest major version directory was along the lines of `v2.x/` rather than `2.x/`.

That way, if someone was looking at a directory listing of all release folders, those ones will appear alongside the individual `v2.x.y` release folders

Issue #239, opened 2019-04-29
Minify CSS output


james-nash:
**Is your feature request related to a problem? Please describe.**
Currently, Gravity's pre-compiled CSS is not minified at all. It wasn't an issue in the past, because our only consumer - the Buildit website - compiles the SASS and does its own minification. However, now that we publish to CDN and are beginning to gain more consumers - some of whom may wish to just use the precompiled CSS - it makes sense to provide something more optimised.

**Describe the solution you'd like**
* We should update our Gulp tasks that compiles to SASS to have some sensible minification:
    * I'd recommend PostCSS plugins like:
        * [CSS MqPacker](https://github.com/hail2u/node-css-mqpacker) <-- I've gotten some significant filesize reductions by using this in the past!
        * [CSS Nano](https://cssnano.co/)
* We should enable sourcemaps, to allow browser dev tools to map things back to the source SASS files.

-----
Related to #238

PR #249, opened 2019-05-07
perf: ⚡️ adds CSS minification

c1rrus:
✅ Closes: #239

**Description**
Adds the following [PostCSS](https://postcss.org/) plug-ins to the SASS compilation task:
* [CSS MQ Packer](https://github.com/hail2u/node-css-mqpacker): De-dupes media query blocks by moving rule sets that are within identical media query blocks, into a single media query block. Since we often use media queries in per-component SASS files, this alone removes a fair chunk of redundant CSS - 9.2% to be precise. :warning: **Note:** Unfortunately, this NPM packages is longer maintained and has therefore been deprecated. I couldn't find an equivalent alternative though and it seems to work, so I've left it in for the time being.
* [CSSNano](https://cssnano.co/): Does a whole bunch of minification stuff

Based on some spot checks, this does not appear to have broken our CSS in any way.

Here are the before and after file sizes of `dist/gravity.css`:

```
No minification:         23573 bytes

CSS MQ packer only:     21404 bytes (9.2% reduction)

CSSNano only:           16669 bytes (29.3% reduction)

both:                   14845 bytes (37% reduction)
```


c1rrus:
It looks like another developer intends to fork and revive CSS MQ Packer! :tada: 
https://github.com/simbo/css-mqpacker-sort-mediaqueries/issues/1

We can migrate over to that version once it's ready.

PR #250, opened 2019-05-08
Basic Card component

### LINKED ISSUE #174: Add card component(s)
A UI inventory of https://www.builditonboarding.com/, which intends to migrate to Gravity, highlighted the need for a card(-like) components which we don't currently have.

Essentially, we need something that is a link to an article or post within a site and contains:

* A featured image (could be optional)
* The title of the thing being linked to
* A short blurb about or excerpt from the thing being linked to (optional?)

When we implement this, we should take a close look at this excellent article: https://inclusive-components.design/cards/ to ensure we choose the most appropriate mark-up.

dw-buildit:
Closes #174

**Description**
Adds a basic card component with component specific styles updated to match the new website design.


packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -0,0 +1,46 @@
+.grav-c-list-cards-basic {
+  $breakpoint-1: gravy-breakpoint(medium);
+  $breakpoint-2: gravy-breakpoint(large);
```
james-nash: `$breakpoint-2` doesn't seem to be used. Remove it?

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -0,0 +1,46 @@
...
+  padding: 0;
+  list-style-type: none;
+
+  @media all and (min-width: $breakpoint-1) {
```
james-nash: Perhaps merge this `@media` block with the one below. That way we have all the non-MQ stuff first, followed by all the inside-MQ stuff. Makes it easier to read IMHO.

In this case you might as well ditch the `$breakpoint-2` var, since you'll only be used `gravy-breakpoint(medium)` once.

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -0,0 +1,46 @@
...
+  > li + li {
+    margin-top: $grav-sp-xl;
+  }
+
+
+  @media all and (min-width: $breakpoint-1) {
+    > {
+      // 2 col layout
+      li {
+        display: flex;
```
james-nash: I think this `display: flex;` is redundant. The `<li>`s are the flex _items_, no?

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -0,0 +1,46 @@
...
+        flex: 50%;
+        padding: 0 $grav-sp-l;
```
james-nash: Hmm... I assume this horizontal padding is to create the gutter between adjacent cards. Trouble is, it also creates a visible inset along the left and right edges of the list, when in 2 coloumn layout.

I suggest removing this padding and instead changing the flex basis above to something like `calc( (100% - #{$grav-sp-xl}) / 2);`.

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -0,0 +1,46 @@
...
+      }
+
+      // Remove top margin from 2nd card
+      // since it appears at top of list now
+      li:nth-child(2) {
+        margin-top: 0;
+      }
+
+      // Callouts go full-width
+      .grav-c-list-cards-basic-callout {
```
james-nash: Hmm.. given that we use BEM-ish conventions, I'd say this callout class ought to be an element modifier

I'd argue the `<li>` is an _element_ of the `.grav-c-list-cards-basic` _block_, and this is a _modifier_ of that element.

So, even though it's a bit verbose, maybe `.grav-c-list-cards-basic--item__callout` is more correct.

Alternatively, I suppose you could argue the "callout" is an _element_ of the list (just one that is not always used) and there make the class `.grav-c-list-cards-basic--callout`.

I'll let you choose which you prefer.

is either an element (the `<li>`s it's applied to are after all _elements_ of the `.grav-c-list-cards` _block_) or a modifier (as in, it _modifies_

PR #253, opened 2019-05-09
New typography styles (WEB-288)

james-nash:
**Description**
Implements the new typography styles from the Buildit website redesign.

* Replaces Google Fonts include with SASS mixins that generate `@font-face` rules to pull in our CDN-hosted copies of Wipro Akkurat
    * These also add `local()` sources, so that if a user happens to have "Wipro Akkurat" installed, it will use that instead of downloading font files. I tested this in Firefox, Safari and Chrome and it behaves as intended
    * The CDN font files have cache headers set such that browsers will typically not revalidate or redownload the files for a very long time. This means, after the first visit, those files should be cached by users' browsers and on subsequent page visits always get loaded from the browser's cache
* Removes `@buildit/gravy` SASS lib
    * Primarily this was done in order to be able to implement responsive text sizes (more on that below)
    * Other stuff we used Gravy for, such as breakpoints, have now been replicated in Gravity's own SASS code
    * Arguably, this makes future Gravity dev work a tad easier since we don't need to look at (or modify) and external library for some of our core styles
* Removes `normalize-css` SASS lib
    * It turns out this had been dormant all along! Removing this makes no difference to our CSS output.
* Implements new typographic scale
    * Using the `modularscale` SASS lib (which were already using), you can create fluid text sizes, which increases linearly between 2 breakpoints but are fixes below or above those breakpoints. To avoid excessive amounts of `calc()` and media queries in our CSS output, I create a set of custom properties for various text sizes.
   * For convenience, SASS functions and a mixin are provided and I have updated existing code to use those (instead of the now-removed `gravy`/`typi` counterparts).
   * I additionally generate a set of utility classes, so that downstream consumers can easily apply a given font size to an element.


packages/gravity-ui-nunjucks/src/_patterns/00-particles/01-typography/00-fonts.njk
```diff
@@ -1,16 +1,18 @@
 {% for font in fonts %}
-  <h4 class="sg-pattern-sub-heading">{{font.name}}</h4>
+  <h4 class="sg-pattern-sub-heading">{{font.cssFontFamily}}</h4>
   <ul class="sg-fonts">
     {% for weight in font.weights %}
       <li class="sg-fonts__font">
         <strong class="sg-fonts__weight">{{weight.name}} weight</strong>
-        <code class="sg-fonts__css">font-family: {{font.cssFontFamily}}; font-weight: {{weight.cssFontWeight}}; font-style: normal;</code>
+        <span class="sg-fonts__css"><strong>CSS</strong>: <code>font-family: {{font.cssFontFamily}}; font-weight: {{weight.cssFontWeight}}; font-style: normal;</code></span>
+        <span class="sg-fonts__css"><strong>SCSS</strong>: <code>@include gravy-font({{font.name}}, {{weight.gravityWeight}});</code></span>
```
dw-buildit: I thought `gravy` was out? Is this something different or can it be removed?

packages/gravity-ui-nunjucks/src/_patterns/00-particles/01-typography/01-text-sizes.njk
```diff
@@ -0,0 +1,7 @@
+<dl>
+{% for textSize in textSizes %}
+  <dt><strong>CSS util class</strong>: <code>.grav-u-font-size-{{ textSize.classSuffix }}</code></dt>
+  <dt><strong>SASS mixin</strong>: <code>@include grav-font-size({{ textSize.scale }});</code></dt>
```
dw-buildit: Same question about gravy as the sass mixin

packages/gravity-ui-web/src/sass/00-settings/_breakpoints.scss
```diff
@@ -0,0 +1,10 @@
+$grav-breakpoint-smallest: 32rem;
+$grav-breakpoint-largest: 75rem;
+
+/// Major layout breakpoints
+$grav-breakpoints: (
+  small: $grav-breakpoint-smallest,
+  medium: 44rem,
+  large: 56rem,
```
dw-buildit: What's the thinking about having variables for smallest and largest breakpoints but not for the medium/large?

Is there no situation where you'd need those? Just thinking out loud.

packages/gravity-ui-web/src/sass/00-settings/_modularscale-settings.scss
```diff
@@ -1,7 +1,25 @@
...
+  ),
+
+  $grav-breakpoint-largest: (
+    ratio: $minor-third,
+    base: $grav-font-size-largest-viewport,
+  )
 );
+
+$grav-scale-min: -2;
+$grav-scale-max: 6;
```
dw-buildit: There's probably some good reason this is -2 to 6...?

packages/gravity-ui-web/src/sass/00-settings/_typography.scss
```diff
@@ -1,8 +1,97 @@
...
+
+$grav-font-src-path: 'https://static.buildit.digital/fonts/v1.0.0';
+
+
+/**
+  Map of available typefaces.
+*/
+$grav-typefaces: (
+  primary: (
+    stack: ($grav-font-family-primary, 'Akkurat', 'Arial', system-ui, sans-serif),
```
dw-buildit: Is 'Arial' as a fallback from our brand guidelines? I wondered if Helvetica is closer to Akkurat (very limited knowledge on the subject of fonts though).

packages/gravity-ui-web/src/sass/02-generic/_typography.scss
```diff
@@ -0,0 +1,33 @@
...
+  line-height: inherit;
+
+  // Text rendering tweaks (formerly came from Gravy lib)
+  text-rendering: optimizeLegibility;
+}
+
+
+:root {
+  // Generate a set of CSS custom properties for our
+  // scale of repsonsive font-sizes. These will all
```
dw-buildit: minoristest of typos re**ps**onsive 🤦‍♂

packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
```diff
@@ -1,22 +1,21 @@
-$grav-page-breakpoint-medium: gravy-breakpoint(medium);
```
dw-buildit: You've removed this variable, but then it's declared on line 18 anyway. Maybe we could remove that declaration too, or use it if it seems appropriate.
```
.grav-c-page-footer__lists {
  $grav-page-breakpoint-medium: grav-breakpoint(medium);
```

packages/gravity-ui-nunjucks/src/_patterns/00-particles/01-typography/00-fonts.njk
james-nash: Ah! Nope, I just forgot to update that sample code. You're quite right, the `gravy-font()` mixin no longer exists.

I'll push a fix for that soon

packages/gravity-ui-nunjucks/src/_patterns/00-particles/01-typography/01-text-sizes.njk
james-nash: No Gravy here. This is `grav-...` (no `y`) :-)

packages/gravity-ui-web/src/sass/00-settings/_breakpoints.scss
james-nash: Yeah, I just made those variables because I wanted a convenient way to reference those same values elsewhere.

I didn't have the same need for the other breakpoint widths, so I left them as is.

packages/gravity-ui-web/src/sass/00-settings/_typography.scss
james-nash: That's what the Wipro brand book says ¯\_(ツ)_/¯

packages/gravity-ui-web/src/sass/00-settings/_modularscale-settings.scss
james-nash: Yeah, I ought to add a comment.

It's based off @patrickbower's website concept designs. At the largest viewport width, the "minor thirds" scale, gives us font sizes that are pretty close the ones in the design. It just so happened that `-2` matchers the smallest size and `+6` the largest size that occured in the design.

I made them vars though, so that we can easily add _even_ bigger or smaller text sizes, should we ever need them.

packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
james-nash: Yeah, I removed the top ones since that's a global variable and probably ought not to be there. I left the second one because it's within the .grav-c-page-footer__lists so it's scoped to that block. Also it's being used twice so that seemed like an reasonable use of a SASS var.

In hindsight though, it might be neater to just use grav-breakpoint(medium) throughout.

Happy to make that change if you agree.

packages/gravity-ui-web/src/sass/03-elements/_headings.scss
```diff
@@ -4,11 +4,30 @@ h3,
...
+
+h1 {
+  @include grav-font-size(6);
+}
+
+h2 {
+  @include grav-font-size(5);
+}
+
+
```
ashblue: Might want to include a rule to remove redundant whitespace.

packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
dw-buildit: for me, i'd prefer the sleek looks of `grav-breakpoint(medium)`

james-nash: Your wish is my command

packages/gravity-ui-web/src/sass/03-elements/_headings.scss
james-nash: Yup. I've manually fixed this like a savage for now though.

PR #257, opened 2019-05-14
feat(basic card): added hover colour and svg icon treatment

dw-buildit:
affects: @buildit/gravity-ui-nunjucks, @buildit/gravity-ui-web

**Description**
Adds rollover (hover) effects to the basic card component when it is being used with a link.

### DIFF
--- a/packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
+++ b/packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
@@ -1,4 +1,5 @@
 .grav-c-card-basic {
+  $card-image-height: 16rem;
   width: 100%;
 
   h1,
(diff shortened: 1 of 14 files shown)


packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
```diff
@@ -16,12 +17,55 @@
...
+
+      width: 4rem;
+      height: $card-image-height;
+      position: absolute;
+      top: 0;
+      left: 50%;
+      margin: 0;
+      opacity: 0;
+      transform: translate(-50%, 0);
+      transition: opacity 0.1s ease-in-out;
```
james-nash: Would it make sense to have local SASS vars for the transition duration and timing function?

If they were ever changed, there's less risk of the overlay and arrow accidentally drifting out of sync.

Feel free to ignore though. I'm probably overthinking / overengineering as usual :-P

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -1,5 +1,5 @@
 .grav-c-list-cards-basic {
-  margin: 0 -$grav-sp-l;
+  margin: 0;
```
james-nash: As per our conventions, we should avoid setting explicit margins "outside" our components as this intereferes with the global `* + *` rule we have to add a margin between successive, stacked blocks.

I realise now that this was happening even before this change. I missed it in the last review. Sorry.

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -10,6 +10,7 @@
   @media all and (min-width: gravy-breakpoint(medium)) {
     display: flex;
     flex-wrap: wrap;
+    margin: 0 -$grav-sp-l;
```
james-nash: So, since we _do_ want the negative horizontal margins, we can bend our rules here. However, let's _not_ set top and bottom margins. That way the automatic vertical spacing will still work.

I just tried and `margin: inherit -$grav-sp-l;` doesn't seem to work, so I'm afraid you'll need to ditch the shorthand form and do:
```
margin-left: -$grav-sp-l;
margin-right: -$grav-sp-l;
```

packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
dw-buildit: We could define transitions and timings at a different level in gravity. That might be nice, and then just use a mixin to apply them.

(Not necessarily to keep the arrow and background transitions in sync, as I can see a future need for them to be out of sync on purpose.)

@james-nash maybe i could create a `tools/_transitions.scss` and put mixins in there (and for now, the variables they use). Does that sound right to you?

dw-buildit:
@james-nash i've added the transition mixin along with updating the transition speeds and using the mixin where appropriate throughout.

dw-buildit:
I've marked this as WIP again, as now the typography is merged there's a few changes to make.

PR #259, opened 2019-05-14
File size history

### LINKED ISSUE #238: Set and track a performance budget
So far we have not really paid attention to the file size of our CSS, JS and SVG output (and possible font files in future). As Gravity grows, we really ought to though.

We should set ourselves a performance budget (I think for now just some reasonable file-size thresholds would suffice) and then add some tests to our build that check the output file sizes and error if they exceed the limits we've set ourselves (maybe they can also print warnings if we are getting close to our thresholds).

That way we'll get an early warning that our filesize(s) are getting to big and can then make a decision:
* Optimise the code to shrink it down
* Deprecate and remove rarely used things
* Increase our limits (if we agree that everything in the library is needed and cannot be compressed further)

We should also log this over time (so, maybe having an option to output file sizes somewhere like a JSON file would be good). That way we can one-day make nice graphs showing how our file size has changed over time, add notes to releases about how much bigger (or smaller) the library is versus the last release, etc. etc.

Further reading:
* https://addyosmani.com/blog/performance-budgets/
* https://adactio.com/journal/14355
* http://www.performancebudget.io/

thehammeranderson:
Closes #238

**Description**
tracks and reports file size history for develop and master branches


thehammeranderson:
this will keep track of file size history in develop and master (for now) in separate files that will be retrieved and stored on the cdn s3.  There is also a rudimentary report page that will live with the history files on the cdn that currently uses Google Charts to create a multi-series line chart that can be switched between branch histories.

scripts/track_file_sizes.sh
```diff
@@ -0,0 +1,37 @@
...
+gravityDir="packages/gravity-ui-web"
+distDir="$gravityDir/dist/ui-lib"
+cdnGravityDir="gravity-ui-web"
+historyFilename="$TRAVIS_BRANCH-file-size-history.js"
+historyFile="$gravityDir/$historyFilename"
+currentSizesFile="$gravityDir/currentfilesizes.json"
+reportPageName="filesizehistory.html"
+reportPage="$gravityDir/$reportPageName"
+
+if [ ! -z "$TRAVIS_COMMIT" ] && [[ "develop" == "$TRAVIS_BRANCH" || "master" == "$TRAVIS_BRANCH" ]]
```
dw-buildit: should we just do this on master?

would a build of either develop or master would overwrite the same file? or does it create separate files for each branch?

packages/gravity-ui-web/filesizehistory.html
```diff
@@ -0,0 +1,95 @@
+<html>
+
+<head>
+    <link rel="stylesheet" href="https://static.buildit.digital/gravity-ui-web/latest/gravity.css">
+    <script type="text/javascript" src="develop-file-size-history.js"></script>
+    <script type="text/javascript" src="master-file-size-history.js"></script>
+    <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
+    <script type="text/javascript">
```
ashblue: Would probably be best practice to throw this JS into a separate file.

packages/gravity-ui-web/package.json
```diff
@@ -29,7 +29,7 @@
     "build": "gulp",
     "styleguide": "gulp styleguide",
     "start": "gulp serve",
-    "test": "npm run lint && bundlesize",
+    "test": "npm run lint && npm run build && bundlesize",
```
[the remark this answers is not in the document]
thehammeranderson: Calling using the test phase was intentional so the build would fail and give feedback from Travis to Github.

scripts/track_file_sizes.sh
thehammeranderson: each branch now and in the future have their own tracking files

packages/gravity-ui-web/package.json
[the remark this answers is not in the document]
dw-buildit: I just meant it might be nice to be able to call it separately as well as in the test phase. So something like:

```
  "start": "gulp serve",
  "bundlesize": "npm run build && bundlesize",
  "test": "npm run lint && npm run bundlesize",
```

but this is not something that would prevent merging, just a comment.

cdn-index.template
```diff
@@ -1,5 +1,5 @@
 <html>
-<head><link rel="stylesheet" href="https://static.buildit.digital/gravity/v1.0.1/gravity.css"></head>
+<head><link rel="stylesheet" href="https://static.buildit.digital/gravity-ui-web/latest/gravity.css"></head>
```
james-nash: 👍

.travis.yml
```diff
@@ -15,6 +15,8 @@ script:
   - npm run test
 before_deploy:
   - pip install --user awscli
+  - npm install -g file-size-report
```
james-nash: Would it not be better to add `file-size-report` as a `devDependency` of the root `package.json`?

cdn-index.template
```diff
@@ -1,5 +1,5 @@
...
 <body>
 <h1>Versions</h1>
 <ul>{versionlist}</ul>
```
james-nash: Can you please add a link to the file size history page on here. That'll make it easier to find / discover later on. Just chuck it at the bottom of the page for now ;-)

packages/gravity-ui-web/filesizehistory.html
james-nash: While I agree, I'd say this is low priority for now.

I think it would be nice to spend a bit of time polishing the layout and design of this page and the versions page. If we did that, it could also be an opportunity to tidy up JS.

But, I'd suggest that's done as a future enhancement and doens't block this PR.

PR #261, opened 2019-05-15
feat(articles): new related items component

ashblue:
New related items component.


packages/gravity-ui-web/src/sass/05-components/03-organisms/03-articles/_00-related-items.scss
```diff
@@ -0,0 +1,17 @@
+.grav-c-related-items {
+  max-width: 540px;
+  margin: 0 auto;
```
james-nash: As per Gravity's layout conventions, components shouldn't set any (max-)width or outer margins on the component itself, so we should remove these lines.

james-nash: If we're going with the approach of turning this into a list, you can add equivalent styles to the `<li>`s within.

Please try to use `rem` rather than `px` though.

ashblue: @james-nash Without this the text block gets extremely long and hard to read. Are you sure we want to remove the max width?

Screen Shot 2019-05-17 at 2 09 57 PM

ashblue:
@james-nash I think this is good now. Just need your feedback on this item https://github.com/buildit/gravity-ui-web/pull/261#discussion_r285272894

packages/gravity-ui-web/src/sass/05-components/03-organisms/03-articles/_00-related-items.scss
```diff
@@ -0,0 +1,24 @@
...
+  max-width: 540px;
+  margin-right: auto;
+  margin-left: auto;
+  padding: 0;
+  list-style: none;
+  text-align: center;
+
+  > li {
+    span {
+      @include grav-color-grp-b-apply('color', 'neutral-alt');
```
james-nash: `neutral-alt` does not exist amongst the group B colors, what you want here is `neutral-subtle`.

packages/gravity-ui-web/src/sass/05-components/03-organisms/03-articles/_00-related-items.scss
```diff
@@ -0,0 +1,24 @@
+.grav-c-related-items {
+  max-width: 540px;
```
james-nash: I'd move the `max-width`, `margin-right` and `margin-left` rules down into the `> li { ... }` block.

As per our layout conventions, components shouldn't set explicit sizes on themselves, but they are welcome to do so on their own children.

Also, if we ever enhance this in the future to show successive links side-by-side, it'll be easier since we just need to layout the `li`s and don't need to fiddle with the enclosing `ul`'s size.

Issue #178, opened 2019-01-25
Add components for article meta-data


james-nash:
The architecture microsite, design systems glossary and possible additions to the main Buildit website such as case-studies or blog posts will require UI components to display the typical meta-data associated with that kind of content:

* Author
* Post date / last modified date
* Tags
* Categories

PR #263, opened 2019-05-17
Feature/next/article

dw-buildit:
This PR is two components and some global style updates.

The two components are:

**Page Intro**
To be used as the `<heading>` for most of our pages. It includes the heading, description and 
optional time and image as per reference designs.

**Ostentatious Copy**
This is expected to contain copy that you want to look a little more showy. Generally adds extra padding between elements and around the component itself. Pulls pull quotes out (margin wise) and on larger screen sizes indents a lot from the left.

**Style Updates**
The PR also includes updates to many text element, list element and some padding styles. Again to better reflect the reference design.

**Also**
There's styling for the `<hr>` element.
There's a page for viewing all the article components with semi-realistic data.


ashblue:
For the 2nd commit `feat(notch decoration): added parameters to the mixin` is that really a feature? I feel like it should fall under a chore or refactor because it isn't declaring anything new for the repository.

james-nash:
Just adding a cross reference to issue #178, since this partially covers that.

Partially, since we don't yet have provisions for article author, tags and/or categories. However, all of those could be added in a future PR, after the website MVP is complete.

dw-buildit:
> For the 2nd commit `feat(notch decoration): added parameters to the mixin` is that really a feature? I feel like it should fall under a chore or refactor because it isn't declaring anything new for the repository.

I did think that this was a feature that developers of Gravity would be able to use (dynamic notch size/position) although I wouldn't actually expect any users to use it. I've changed this commit's message anyway 👍

packages/gravity-ui-nunjucks/src/_patterns/03-organisms/00-page-structure/04-page-intro.md
```diff
@@ -0,0 +1,5 @@
+## Purpose
+
+This is the component for use as the page heading.
+
+It consists of a `<h1>` as the page title, a `<p>` for a short description of the page, an optional `<time>` to display a date related to this content (usually publication date), and an optional `<img>` which will automatically gain a downward pointing notch.
```
james-nash: Maybe add some guidance here about when to use a `<header>` element or a `<div>` to markup the page intro.

packages/gravity-ui-nunjucks/src/_patterns/05-pages/05-article.njk
```diff
@@ -0,0 +1,29 @@
...
+    <header class="grav-c-page-intro">
+      {{ pageIntroContents(article.title, article.description, article.dateTime, article.img) | safe }}
+    </header>
+    <div class="grav-c-ostentatious-copy">
+      {{ article.copy | safe }}
+    </div>
+  </article>
+  <hr />
+  <aside class="grav-o-container">
+    <ul class="grav-c-related-items">
```
james-nash: You should replace this with `{% include 'organisms-related-items' %}`. That way, if the related items component ever changes, it will be reflected here. Also, it lets Pattern Lab / Fractal create lineage links between those components in the pattern library.

packages/gravity-ui-web/src/sass/05-components/03-organisms/_copy-ostentatious.scss
```diff
@@ -0,0 +1,38 @@
...
+    }
+  }
+
+  img {
+    width: 100%;
+    height: auto;
+  }
+
+  @media all and (max-width: grav-breakpoint(medium)) {
+    padding-right: $grav-sp-s;
```
james-nash: We can avoid this `max-width` media query. You can set this padding and the blockquote styles outside of the media query.

Then, below, in the `min-width` media query, where you already alter the blockquote styles as needed, you can update the padding to be large instead of small.

packages/gravity-ui-nunjucks/src/_patterns/05-pages/05-article.njk
dw-buildit: Definitely! That was an oversight.

dw-buildit:
@james-nash CR amends completed.

PR #273, opened 2019-05-21
Color updates

james-nash:
Adds a number of color-related tweaks and updates:

* Upgrades `@buildit/gravity-particles` to newer v0.3 release and itegrates the updates color schemes it provides
* Removes some (now no longer needed) color overrides to nav links
* Adds support for OS's "dark mode" setting in the page's default color scheme and the new `grav-u-color-scheme-contrast` color scheme utility. (Other color schemes continue to just apply the same colors regardless of the user's "dark mode" settings)
* Updates the styling of `<mark>` elements, which were not integrated with the color system and could therefore become unreadable when certain schemes were applied.
* Applies the "contrast" color scheme to page footers in the pattern library to demo it


packages/gravity-ui-nunjucks/gulp/color-scheme-tables.js
```diff
@@ -18,7 +19,7 @@ function wcagRating(contrastRatio) {
 }
 
 function gravityCssCustomPropName(group, colorName) {
-  return `--grav-co-grp-${group}-${kebabCase(colorName)}`;
+  return `--grav-co-grp-${group}-${kebabCase(startCase(colorName))}`;
```
dw-buildit: Is there a benefit of chaining `kebabCase()` and `startCase()` here?

packages/gravity-ui-nunjucks/src/_patterns/00-particles/00-color/02-color-schemes.njk
```diff
@@ -50,10 +50,10 @@
 </table>
 {% endmacro -%}
 
-{% for schemeName, schemeTableData in colorSchemes %}
-<section class="sg-color-scheme grav-u-color-scheme-{{ schemeName }}">
-  <h4>{{ schemeName | title }} color scheme</h4>
-  <p>Can be applied to containers using the <code>grav-u-color-scheme-{{ schemeName }}</code> utility class.</p>
-  {{ colorSchemeTable(schemeTableData) | safe }}
+{% for schemeKey, schemeData in colorSchemes %}
```
dw-buildit: I don't think you need `schemeKey` here

packages/gravity-ui-nunjucks/src/_patterns/03-organisms/00-page-structure/01-page-footer.njk
```diff
@@ -1,7 +1,7 @@
 {%- from 'atoms-icon' import icon -%}
 {%- from 'atoms-image' import image -%}
 
-<footer>
+<footer {% if footer.class %} class="{{ footer.class }}"{% endif %}>
```
dw-buildit: Unless this is a linting issue you can get rid of one of the spaces here, either inside the `if` or before it.

packages/gravity-ui-web/src/sass/00-settings/_color-schemes.scss
```diff
@@ -8,4 +8,10 @@
 @import 'gravity-particles/color-schemes';
 
 /// The name of the default color scheme that is applied to pages
-$grav-co-scheme-default-name: 'light';
+$grav-co-scheme-default-name: 'teal-white';
+$grav-co-scheme-default-dark-name: 'teal-dark-gray';
+
+
```
dw-buildit: double line space 😱

james-nash: Forgive me for my sins!

packages/gravity-ui-nunjucks/src/_patterns/00-particles/00-color/02-color-schemes.njk
james-nash: I had to because `colorSchemes` was an object, so I needed to iterate over `key, value` pairs. Removing `schemeKey` (even though it's not used in this case) would make it only iterate over the keys.

However, since I'm not using those keys anymore, I've now change the JS that generates the data to output an array instead and therefore I can do what you ask :-)

packages/gravity-ui-nunjucks/gulp/color-scheme-tables.js
james-nash: Nope. I've killed it now

I _think_ I was having trouble with some names getting mangled and that's why I originally did this. However, it seems to work fine without now.

packages/gravity-ui-nunjucks/package.json
```diff
@@ -44,12 +44,12 @@
     "access": "public"
   },
   "devDependencies": {
-    "@buildit/gravity-particles": "^0.2.0",
-    "@buildit/gravity-ui-web": "^3.0.0-alpha.0",
+    "@buildit/gravity-particles": "^0.3.0",
+    "@buildit/gravity-ui-web": "^3.0.0-alpha.1",
```
dw-buildit: this should be `alpha.2`, don't know if this was a merge conflict when you merged `next` in

packages/gravity-ui-nunjucks/package.json
```diff
@@ -59,11 +59,12 @@
...
     "gulp-sass": "^4.0.2",
     "lodash.kebabcase": "^4.1.1",
+    "lodash.startcase": "^4.4.0",
     "minimist": "^1.2.0",
-    "semantic-release": "^15.13.13"
+    "semantic-release": "^15.13.14"
   },
   "peerDependencies": {
-    "@buildit/gravity-ui-web": "^2.0.0",
+    "@buildit/gravity-ui-web": "^3.0.0-alpha.1",
```
dw-buildit: `alpha-2` again

PR #287, opened 2019-05-29
Upgrade travis python

thehammeranderson:
**Description**
upgrade python on travis to newest version to get rid of SNI SSL issues with aws install and calls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@ script:
   - npm run-script styleguide
   - npm run test
 before_deploy:
-  - sudo apt-get -y install python3
-  - sudo apt-get -y install python3-pip
+  - pyenv global 3.6.1
   - pip3 install --user awscli
   - ./scripts/track_file_sizes.sh
 deploy:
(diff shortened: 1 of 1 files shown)


thehammeranderson:
> This looks fine, but I wonder if there's a more elegant solution.
> 
> Looking at the logs of a recent build, it seems the Travis VMs use Ubuntu Trusty ([and that's likely to change to Xenial soon](https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476)). According to the Travis docs, [it comes with both python 2.7.13 & 3.6.1](https://docs.travis-ci.com/user/reference/trusty/#python-images) and also `pyenv`. It defaults to python 2.7.13 though.
> 
> So, as suggested in this [Stackoverflow answer](https://stackoverflow.com/a/55096093), we should be able to just use `pyenv global 3.6` to switch to the already-installed python 3.6.1.
> 
> Could you please try that? If it works, we'll save ourselves a redundant download and install. :-)

I can.  When I looked at the build env logs it looked like it wasn't installed (-) next to it.

PR #292, opened 2019-05-31
feat(callout): new callout organism component

ashblue:
New callout component.


packages/gravity-ui-web/src/sass/05-components/03-organisms/_callout.scss
```diff
@@ -0,0 +1,44 @@
+.grav-c-callout {
+  position: relative;
+
+  p {
+    font-size: 1.5rem;
```
james-nash: For font sizes, please use one of the pre-defined ones we have - there's a scale from `-2` (smallest) to `6` (biggest), with `0` being the "base" size that's used as the default font size on the page. These are all responsive too, so they will increase relative to the viewport with, meaning you don't need that media query below to adjust the font size.

Looking at the designs, I think size `2` is a good fit - the card titles, which are slightly smaller, use `1` so I'm thinking one notch up from that is about right. :-)

To make things a bit easier, we have a mixin that sets the font-size using our scale, so you can replace this line (and the 2nd media query below) with this:
```sass
@include grav-font-size(2);
```

packages/gravity-ui-nunjucks/src/_patterns/03-organisms/00-page-structure/05-callout.njk
```diff
@@ -0,0 +1,10 @@
+{%- from 'molecules-three-column-list-justified' import threeColumnListJustified -%}
+
+<div class="grav-c-callout">
+  <img src="{{ backgroundImage }}" />
```
james-nash: Hmm.. I'm torn about this. Using an `img` element as you've done is fine and works. But, I wonder if using a background image here would be easier. @dw-buildit - do you have a preference?

That being said, _if_ we leave this as an `img`, then you must add an `alt=""` attribute to it. Alt attributes are mandatory but, if the image is purely decorative, it's value should be an empty string. See: https://www.w3.org/WAI/tutorials/images/decision-tree/

dw-buildit:
Thanks @ashblue.

I have a question about this component's purpose and whether it fits in to gravity.

The task to create this component we initially categorised as website specific (the ticket is in the website component epic rather than gravity component) so my assumption would be that the code would be within the website codebase only. The way you've built it, however, could act as a reusable component and therefore could exist as a gravity component. I have a reservation though.

Gravity has been trying to move towards being opinionated on accessibility including text colour contrast ratio. For any component that has a background as an image it is going to be difficult to enforce that unless the image(s) that can be used are also prescribed by Gravity.

So if we want a component (like this one) with text over a customisable background image we will have to either:
- assume that our consumers are going to use a color scheme and imagery that has a contrast ratio meeting gravity's standards
- provide a set of images in gravity that our consumers can use. The images would meet the contrast ratio and would enforce a contrasty color scheme based on the image chosen
- not allow gravity components to have text over background images - ie make this a rule for this component and future gravity components

Am i missing some way that we can ensure contrast? Or are there any other thoughts on this?
Whatever we decide could be adjusted in future but this seems like a good time to open a discussion.

packages/gravity-ui-nunjucks/src/_patterns/03-organisms/00-page-structure/05-callout.njk
dw-buildit: I think the general rule i've used is use `` for content images and `background-image:` for non content images. The main difference I think is that images get referenced by assistive tech (although you've mentioned how to correctly mark up decorative images) and they get printed. Whereas background images do not.

We don't have a print stylesheet so the printing question is not as clear cut. If we did I'd probably argue that this should be a background image so that we can print the text as standard black on white rather than having a large block print with white text within.

ashblue: I'm honestly fine with using this as an `` or background image. @james-nash and @dw-buildit if you could let me know how to proceed, as I'm not sure how to resolve this atm.

ashblue:
@dw-buildit I think that's a good question but probably way beyond my knowledge of style guides and CSS systems. I'd like to merge this PR in to keep things rolling so it might be better to move that question to an open issue so it doesn't get lost. Simplest answer I can think of is making sure that we add a background color to images to ensure contrast. But maybe that isn't the correct answer.

packages/gravity-ui-nunjucks/src/_patterns/03-organisms/00-page-structure/05-callout.njk
james-nash: Considering both of @dw-buildit's comments, I propose the following:

- We completely remove images from this Gravity component. It would therefore only concern itself with laying out the big text and the 3 column list thing. This side-steps any potential a11y issues for whatever `@buildit/gravity-ui-web` provides out of the box
- Then, in the website's own CSS, code we customise the instance of this component on the homepage by adding the desired `background-image`.

My thinking is, the website project bears then responsibility for choosing an appropriate background image (and color scheme).

Sound good?

ashblue: Sounds good, I'll remove the background image then 👍

PR #302, opened 2019-06-06
Feature/next/jobs and location pages

dw-buildit:
- Created a column layout object level helper mixin and some classes to use it.
- Refactored `three-column-list-justified` and `list-cards-basic` components.
- **BREAKING CHANGE:** removed `two-column-text` and `two-column-block` components.
- Updated use cases of those components in the nunjucks-ui package.
- Created a `n-column-list` helper in nunjucks-ui which uses the new column layout classes.

- The Careers templates have been renamed Jobs in line with the website.

- The Locations template has had rounded corners removed and "View map" text added to the image link.


packages/gravity-ui-web/src/sass/05-components/02-molecules/02-lists/01-two-column-list.scss
```diff
@@ -0,0 +1,26 @@
+.grav-c-two-column-list {
```
james-nash: Hmm.. this doesn't seem much different to the existing `.grav-c-two-columns-block`, which also lays out items in 2 columns using flexbox.

My preference would be to merge the two. That might also be a good time to rename the class to an "object" - e.g. `.grav-o-two-columns` or similar - since it's just a layout thing, not really a _component_.

On the Nunjucks side, we'd then also need to replace the 2 templates with one. The current `molecules-two-columns-block` one is a bit dumb TBH because it hard-codes location cards for the items.

A better approach would be for a `molecules-two-columns` Nunjucks template to define a `{% block %}` that can be extended elsewhere - such as our sample page templates - to contain anything we like.

Come to think of it, maybe we should combine `.grav-c-list-cards-basic` with this too! In terms of layout these are all doing more or less the same. And the feature of stretching an item to be full-width could well be useful beyond cards. Although, for now I'd be content for the two-column ones to be combined and we can look at merging in the cards list later.

dw-buildit: The only difference is the required list styles on the list (such as `list-style: none;`), so I can understand the want to merge them. They do behave slightly different (which might not be a good thing) layout wise.

I'm not sure about combining the behaviour `.grav-c-list-cards-basic` has with all two column layouts is desired but I can understand the potential uses of a more generic class to do that. So having a two column class and a separate two column with full width 'hanging' class makes sense to me.

Perhaps there should be two basic two-column classes too for the different content arrangements:
```
|0|1|    |0|3|
|2|3| or |1|4|
|4|5|    |2|5|
```

james-nash: Sounds sensible.

Will you be in tomorrow? Perhaps we can sit down and hash this out together?

dw-buildit: Changes made to create a two column object level helper class and refactor other components (or remove some). Feel free to re-review @james-nash  (i'm about to update the PR description).

packages/gravity-ui-web/src/sass/01-tools/_layouts.scss
```diff
@@ -0,0 +1,53 @@
...
+  padding-right: 0;
+  padding-left: 0;
+
+  > * {
+    width: calc((100% - #{(($column-count - 1) * 2) * $column-spacing}) / #{$column-count});
+    margin-right: $column-spacing;
+    margin-left: $column-spacing;
+  }
+
+  > *:nth-child(#{$column-count}n) {
```
james-nash: Perhaps add a comment for each of these `:nth-child` selectors to explain what they do. I know because we were chatting about it yesterday, but I'm not sure they'd be totally self-explanatory to others.

Perhaps something like:
```scss
// Remove outer margin from right-most item
> *:nth-child(#{$column-count}n) { ... }

// Remove outer margin from left-most item
> *:nth-child(#{$column-count}n - #{$column-count - 1}) { ... }

// Remove top margin from all items in first row
> *:nth-child(-n + #{$column-count}) { ... }
```

packages/gravity-ui-web/src/sass/01-tools/_layouts.scss
```diff
@@ -0,0 +1,53 @@
+// a mixin to add column based layout styles
+@mixin grav-l-column-layout($column-count: 2, $column-spacing: $grav-sp-m) {
```
james-nash: Hmm.. I think the name `$column-spacing` is slightly ambiguous. I this case it is _half_ the gutter width, but without looking at the code I probably would have assumed this parameter was meant to be the _whole_ gutter, or "gap" if you prefer, width - i.e. the spacing between columns.

When talking about layout grids, "gutter" is a commonly used and fairly well understood term meaning the whole gap between columns (or rows): https://uxdesign.cc/the-anatomy-of-a-grid-c955d5355fae

However, a brief chat with @danbull reminded me that in CSS they tend to use the word "gap" for same purpose. E.g. `column-gap`, `grid-column-gap`, etc. so maybe that's a better name to use rather than "gutter".

So perhaps something along these lines would be more intuitive:

```scss
@mixin grav-l-column-layout($column-count: 2, $column-gap: $grav-sp-l) {
  $column-spacing: $column-gap / 2;
  ...
}
```

packages/gravity-ui-web/src/sass/01-tools/_layouts.scss
```diff
@@ -0,0 +1,53 @@
...
+  list-style: none;
+
+  > li {
+    padding-right: 0;
+    padding-left: 0;
+  }
+}
+
+// a mixin to expand items in column layouts if the final row is not full.
+@mixin grav-l-column-no-danglies {
```
james-nash: Love it! :-D

packages/gravity-ui-web/src/sass/04-objects/_layouts.scss
```diff
@@ -0,0 +1,28 @@
+// some utility classes to use the grav-l-column-layout mixin directly as classes with sensible breakpoints.
+.grav-o-two-column {
+  @media (min-width: grav-breakpoint(small)) {
+    @include grav-l-column-layout(2);
+  }
+}
+
+.grav-o-up-to-four-column {
```
james-nash: Hmm... I think having a consistent naming scheme between these classes would be nice. Either both should be `.grav-o-up-to-X-column` or both should be `.grav-o-X-column`. I'm leaning towards the latter, as its more concise. In any case, I think the documentation in the "n-column" markdown file should explain that these classes are always "up to N columns from the widest breakpoint onwards, and reduce to lower number of columns on narrower viewports".

Also, if we're adding 2 and 4 column classes, should we not add a 3 col one for completeness?

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -1,31 +1,28 @@
...
+  @include grav-l-column-list;
 
   > li + li {
     margin-top: $grav-sp-xl;
   }
 
   @media all and (min-width: grav-breakpoint(large)) {
-    display: flex;
-    flex-wrap: wrap;
+    @include grav-l-column-layout(3, $grav-sp-l);
```
james-nash: Should this not be 2 columns instead of 3?

packages/gravity-ui-web/src/sass/01-tools/_layouts.scss
```diff
@@ -0,0 +1,53 @@
+// a mixin to add column based layout styles
+@mixin grav-l-column-layout($column-count: 2, $column-spacing: $grav-sp-m) {
+  display: flex;
+  flex-wrap: wrap;
+  padding-right: 0;
+  padding-left: 0;
+
+  > * {
```
james-nash: Having seen what you do later on for `.grav-c-list-cards-basic`, I have a proposal: Why not change this mixin to work the same way that does (and thus save yourself a few overrides there later on)? I.e.:

* Set negative left & right half-gutter margins on the container
* Remove the `nth-child()` rules below that zero the outer margins of the left and right-most items

I tried it quickly locally and it appears to work fine.

This has the added bonus that we could make a generic version of `.grav-c-list-cards-basic__item--callout` that can stretch a single item of a "no-danglies" column layout to be full-width. (As it stands, that would throw the `nth-child` logic out of whack since all subsequent items would be off by one)

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
```diff
@@ -1,31 +1,28 @@
...
+    @include grav-l-column-no-danglies;
+
     margin-right: -$grav-sp-l;
```
james-nash: If you tweak the `grav-l-column-layout()` mixin as I suggested above, you can remove most of these rules :-)

packages/gravity-ui-web/src/sass/05-components/03-organisms/_list-cards-basic.scss
dw-buildit: whoops, i was playing around proving that they could work with more columns and must have missed setting this back

dw-buildit:
I'll merge this now as there's a small tweak to make on the amended code @james-nash

PR #310, opened 2019-06-17
feat(column layout): adds mixin for full-width items

james-nash:

The `.grav-c-list-cards-basic` component has an element modifier class that can stretch a single
card to span both columns. The CSS logic behind it might be useful to other column layouts, so this
change extracts and genericises it into a mixin that can be used in conjunction with any "no
danglies" column layout. It also updates the `.grav-c-list-cards.basic` class to use this mixin.

### DIFF
--- a/packages/gravity-ui-web/src/sass/01-tools/_layouts.scss
+++ b/packages/gravity-ui-web/src/sass/01-tools/_layouts.scss
@@ -60,7 +60,7 @@
   // its top margin reinstated. Likewise, subsequent items that
   // would have otherwise been in the first row also need their
   // top margins reinstated.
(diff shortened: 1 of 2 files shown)


james-nash:
@dw-buildit Good spot. I've fixed the issue now. I also noticed that I'd accidentally decreased the horizontal gap between cards, so that's fixed too.

Here's my local vs the `next` staging env side-by-side:
image

..and the same again with the 1st item in the list set to be `.grav-c-list-cards-basic__item--callout`:
image

...and again but this time with the 2nd item having the callout class:
image

james-nash:
Good spot with the redundant margin! Thanks for fixing that.

PR #315, opened 2019-06-20
fix: undoes layout regression caused by previous "fix"

james-nash:
A recent fix to the container mixin incorrectly set the width as 100%, which failed to take into
account the outer margins that are applied on smaller viewports. This cause container components to
horizontally overflow the viewport. This fixes that by replacing 100% with a suitable calc()
expression.

### DIFF
--- a/packages/gravity-ui-web/src/ui-lib/sass/01-tools/_container.scss
+++ b/packages/gravity-ui-web/src/ui-lib/sass/01-tools/_container.scss
@@ -1,6 +1,6 @@
 // a simple mixin to create a container for the page
 @mixin grav-l-container($no-margin: false) {
-  width: 100%;
+  width: calc(100% - #{2 * $grav-page-content-inset});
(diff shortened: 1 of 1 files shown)


packages/gravity-ui-web/src/ui-lib/sass/01-tools/_container.scss
```diff
@@ -1,6 +1,6 @@
 // a simple mixin to create a container for the page
 @mixin grav-l-container($no-margin: false) {
-  width: 100%;
+  width: calc(100% - #{2 * $grav-page-content-inset});
```
dw-buildit: Should this be in the `@if (not $no-margin)` below?

james-nash: Yes, you're right. Will push a fix.

Although, TBH, I don't think that option is used anywhere.

PR #318, opened 2019-06-21
Release v3.0.0-alpha.6

james-nash:
Version bumps from release ` v3.0.0-alpha.6` NPM package.

### DIFF
--- a/packages/gravity-ui-nunjucks/package.json
+++ b/packages/gravity-ui-nunjucks/package.json
@@ -47,7 +47,7 @@
   "devDependencies": {
     "@buildit/eslint-config": "^1.0.0",
     "@buildit/gravity-particles": "^0.3.0",
-    "@buildit/gravity-ui-web": "^3.0.0-alpha.2",
+    "@buildit/gravity-ui-web": "^3.0.0-alpha.6",
     "@buildit/semantic-release": "^1.0.0",
     "@pattern-lab/core": "^3.0.1",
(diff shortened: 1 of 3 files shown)


packages/gravity-ui-nunjucks/package.json
```diff
@@ -47,7 +47,7 @@
   "devDependencies": {
     "@buildit/eslint-config": "^1.0.0",
     "@buildit/gravity-particles": "^0.3.0",
-    "@buildit/gravity-ui-web": "^3.0.0-alpha.2",
+    "@buildit/gravity-ui-web": "^3.0.0-alpha.6",
```
dw-buildit: Is there not an equivalent `package-lock.json` change for the nunjucks package?

james-nash: Nope.

When Lerna does it's "bootstrap" thing, it temporarily removes any local packages dependencies, runs the install and then restores the `package.json`. So, the `package-lock.json` files don't contain refrences to any of the other packages in this monorepo

james-nash: I suspect I didn't really need to update the Nunjucks `package.json` at all, but it felt neater to keep it up-to-date.

Especially later, when we start publishing that to NPM, it'll be important for it's `package.json` to list the correct version of `gravity-ui-web`

dw-buildit: Makes sense, i saw the `gravity-ui-web` lock being updated but of course that's the actual package's version.

PR #323, opened 2019-07-15
refactor(footer): ensure margin above footer

dw-buildit:
Ensure there is the usual vertical margin above the footer, regardless of the content that precedes it.

This should play nicely with preceding containers that specify their own bottom margin as the margins will collapse.

### DIFF
--- a/packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
+++ b/packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
@@ -2,6 +2,7 @@ body > footer {
   @include grav-color-grp-b-apply('border-bottom-color', 'accent');
   @include grav-font-size(-1);
 
+  margin-top: $grav-sp-vertical-gap;
   padding: $grav-sp-l 0;
   border-bottom-width: $grav-st-thickness-medium;
   border-bottom-style: solid;


packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_01-page-footer.scss
```diff
@@ -2,6 +2,7 @@ body > footer {
   @include grav-color-grp-b-apply('border-bottom-color', 'accent');
   @include grav-font-size(-1);
 
+  margin-top: $grav-sp-vertical-gap;
```
james-nash: Hmm...

In `_heydon.scss` we disable the default margin top using:

```scss
...,
body > * {
  margin-top: 0;
}
```

I'm wondering whether we ought to remove that line instead of re-adding a margin for the footer here.

If we did that, page headers would be unaffected (since it's the first child of `body` and therefore has no top margin anyway) and the page footer would get the margin it needs. The only other element being affected would be elements inbetween, but arguably, they _should_ be getting margins between them too.

This would also be more consistent with our component layout guidelines that state individual components should not set explicit margins on themselves.

james-nash: So, I've had a quick peak at the "next" website. I think it wouldn't be _too_ disruptive to go down this route.

We'd probably need to add a util class to zero the top margin (maybe `grav-u-no-margin-top`, or similar?) for the homepage, where we want the hero to appear flush below the header. But, elsewhere stuff should "just work".

dw-buildit: I also had a look last week, but I read your last comment as:
> I think it **would** be _too_ disruptive
🤦‍♂

Will sort it out today.

danbull: I've had a think about this and investigated what other sites do, and I think we might be better off adding padding bottom to the `main` element. This would be important in cases where there is no footer (web apps etc). I think the best way to add the padding would be to introduce a system of utility classes e.g. `.grav-u-pb-small` `.grav-u-pb-large` etc

How do we feel about this?

dw-buildit: I don't mind adding something to the bottom of main but I think using a utility class to do it would mean adding the utility class to pretty much every page of a 'standard' website. So I feel like the utility classes would be secondary to adding the padding or margin as a default.

james-nash: I think some padding utilities could be useful in any case. If you consider a container where someone applies a different color scheme and then puts text inside it, they'd almost certainly want padding above and below so the text. In fact, the new website's "We’re worldwide..." callout is an instance of exactly that kind of thing. Currently it was done as a component, but if we had some padding utility classes, that component would be redundant. It could be replaced by combo along the lines of: `grav-u-color-scheme-contrast grav-o-container grav-u-pb-m grav-u-pt-m`.

That being said, I don't think an automatic bottom padding on `<main>` is desirable. As per our discussion the other day, a subsequent `<footer>` (or any element for that matter) will automatically get a top margin applied by the `* + *` rule (once we remove the current exceptions to that rule).

If we go with another layout change I've proposed in #333, then for simple pages the body would have a margin or padding around all sides, so no footer wouldn't be an issue for simple pages. And, if someone opts-in to the current layout behaviour but chooses not to have a footer, then they can always manually add a padding to `<main>` (or whatever their last element happens to be) via a utility class.

TL;DR: Add some util classes for padding: Yes. Give `main` bottom padding by default: No.

danbull: I'll create a PR for this today. I understand the point about having to add a class to "every" page but in reality it's pretty standard to use some form of templating so it would just be in a few places at most.

james-nash:
I propose closing this PR in favour of the changes proposed in #339. Amonst other things, it removes the `body > *` exception to the lobotomised owl rule, which ensures that page footers _do_ get a standard `margin-top` when they are preceded by other elements.

danbull:
@james-nash great, agree that this should be closed then if @dw-buildit also agrees

PR #324, opened 2019-07-15
refactor(basic card): use neutral alt bg color for slow loading images

dw-buildit:

Adds a subtle background color to the card `img` element so that whilst loading there is a representation to the user of the notch etc.

### DIFF
--- a/packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
+++ b/packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
@@ -16,6 +16,7 @@
   }
 
   img {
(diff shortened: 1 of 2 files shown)


packages/gravity-ui-web/src/sass/05-components/03-organisms/_card-basic.scss
```diff
@@ -16,6 +16,7 @@
   }
 
   img {
+    @include grav-color-grp-a-apply('background', 'neutral-alt', true);
```
james-nash: Nice touch! I like it.

james-nash:
@dw-buildit Would it make sense to do the same thing for page intro images?

PR #330, opened 2019-07-22
feat(layout): sticky page footer is now opt-in

### LINKED ISSUE #133: Flex on body causing undesirable side effects
When `inline` patterns are immediate child of `body` in patternlab they are rendered as `block`. e.g. http://style.buildit.digital/?p=atoms-block-link

I propose maybe adding the `flex` CSS to a class rather than directly on the `body` element.

james-nash:
Closes #133

affects: @buildit/gravity-ui-nunjucks, @buildit/gravity-ui-web

**Description**
Previously, direct descendents of the `<body>` element were laid out such that the page footer would
be flush with the bottom of the viewport on short pages where it would otherwise appear further up.
This was achieved by applying `display: flex;` on the `<body>` element. In some scenarios, this
caused unwelcome side-effects (see issue #133). This change therefore removes this default layout
behaviour and introduces an object class (`.grav-o-sticky-footer`) that consumers can apply to the
`<body>` element to exlicitly opt into the sticky footer layout.

Note, this is a **BREAKING CHANGE**:
Pages relying on the old default sticky footer will now need to add the `.grav-o-sticky-footer`
class to `<body>` in order to continue getting that layout.


james-nash:
Hmmm... I umm and errr about this too.

Perhaps you're right. Maybe we keep it on by default, but add a class for disabling it.

In the Gravity spirit of "if you just write some basic HTML, it should look decent", I think there's a good case for sticky footers by default. But, since we know there are cases where it is unwanted, we should have a standard way of switching it off.

@danbull What do you think?

james-nash:
On a related note, I've been tinkering in CodePen and, regardless of whether it's default or not, would like to change _how_ we implement the sticky footer.

Currently, we turn `<body>` into a flexbox column that is at least as tall as the viewport. It has `justify-content: space-between;` which makes its direct child elements distribute themselves vertically. Finally we set `flex-grow: 1` on `<main>`, which makes it grow to expand any excess space, since it would otherwise be vertically centered between header & footer.

The trouble with this apprach is:
* We're assuming `<main>` is always present - if it is not, things may lay out weirdly
* We're affecting the height of `<main>` - there might be edge cases where the layout of stuff within it could be affected.

So, what I propose is:
* We remove the `justify-content: space-between;` on `<body>`. The default is `flex-start`, which would mean that elements simply stack - just like they would outside of a flexbox. This avoids things spacing themselves out vertically when we don't want them to.
* We remove the `flex-grow: 1;`, so that `<main>`'s height is not altered.
* In order to expand the space between the footer and whatever element preceded it, we can insert an `::after` pseudo element and then use `order` to position it before the footer (I tried it out in a CodePen and it works!). This way it doesn't matter what element comes before the footer.

danbull:
@james-nash I think the focus here really is making the body of the page full height or full height minus the space of the footer below, if there is one. I don't think it's really "sticky". So the thing you'd opt out of is having pages full height. This could be a great chance to experiment with some scss setting vars and some conditionals. e.g. `$grav-auto-expand-content-body: false`. Could be much more graceful than using additional classes/markup.

james-nash:
I agree "sticky footer" perhaps isn't the best name but, alas, that's what I've seen others refer to it as over the years. E.g.: https://css-tricks.com/couple-takes-sticky-footer/ Happy to consider alternatives, but I'm not sure "full height" really does it justice either though. The thing you're trying to affect is the position of the footer, so I think "footer" needs to be part of the name somehow. The (at least) full-height body is merely a means to that end. Maybe "smart footer" (as it has "smart" behaviour and also looks "smart"). Or "footer-at-bottom"? I dunno - I'm open to better suggestions!

Naming aside, I think you're on to something with the settings vars. Customised Gravity builds is something we hadn't really considered at all so far.

Off the top of my head, I'd still be inclined to ship classes that let you explicitly turn this on or off. So, even if the default can be configured in builds, I think projects may still appreciate the ability to occasionally override their chosen default on specific pages. E.g. a site might choose to have "sticky footers" by default, but switch them off for certain pages (maybe because they're marketing pages wirth some highly custom layout or something). Conversely a web app project may opt to disable "sticky footers" by default but want to turn them back on for help pages.

Also, for SPAs that have a structure like:
```
<body>
  <div id="root">
     <header class="grav-c-page-header">...</header>
     ...
     <footer class="grav-c-page-footer">...</footer>
  </div>
</body>
```
...the default which would only apply to `<body>`, wouldn't make any difference one way or the other - they _never_ get a sticky footer, unless we give them a class to switch it on (which they would then apply to `<div id="root">`.

So, putting it all together, how about this:
* We add a SASS var that lets consumers set wether sticky footers are on or off out-of-the-box. (we default it to "on", for the benefit of simple HTML pages)
* We add a class to explicitly disable the sticky footer (which covers projects that use a vanilla Gravity CSS build but need to occasionally switch it off)
* We add a class to explicitly enable the sticky footer (which covers projects that configured their Gravity CSS to be off by default but want to override. It also enables SPAs to apply sticky footer behaviour to containers other than `<body>`.

danbull:
@james-nash sounds good to me and a good framework for future features, e.g. sticky header

- Sensible global default (can be turned off/on via sass var)
- Class to specifically enable/disable default for exceptions

james-nash:
I'm going to close this PR. I have implemented a better and more configurable solution in PR #339.

That PR also contains some other (proposed) layout changes because they're somewhat interdependent. Specifically, the `.grav-o-sticky-footer` class needs to behave differently if it occurs within a `.grav-o-full-bleed` body.

PR #334, opened 2019-07-22
Normalise media queries & update coding standards

james-nash:
**Description**
I noticed that our code had inconsistent styles for writing media queries. Sometimes we'd do `@media all and (min-width: ...)` and other times we'd do `@media (min-width: ...)`.

While mostly harmless (the two are equivalent as far as browsers are concerned), it does add a little unnecessary bloat to our compiled CSS. We use the [`css-mqpacker`](https://github.com/hail2u/node-css-mqpacker) PostCSS pug-in, which will collate blocks of CSS within idential media queries into a single media query block. Since we write media queries in per-component source files for developer convenience, this can lead to a lot of redundant repetition in the CSS output and this plugin helps us avoid that.

However, it considers `@media all and (min-width: ...)` and `@media (min-width: ...)` to be _different_ and so the two are not collated into a single block. Funnily enough, _another_ PostCSS plugin we run _after_ the MQ packer simplifies `@media all and (min-width: ...)` to `@media (min-width: ...)`, so the final output only has that form anyway, but it gets repeated a couple of times.

So, to avoid that, this PR standardises all instances of media queries to be `@media (min-width: ...)`. It also adjust's the plugin's options so that it will always sort `min-width` media queries from narrowest to widest (without this option, some instances were getting jumbled up in the output, which had some unintended consequences for the styling).

While making the changes, I also tweaked the column layout mixin and classes, so that we don't needlessly repeat the same CSS properties at each breakpoint.

Putting it all together has cut down the CSS file size a _little_ bit:

* Before: `30617 bytes` uncompressed filesize. **12** instances of `@media`.
* After: `30048 bytes` uncopressed filesize. **8** insrances of `@media`.
     * ~2% saving in filesize, but 33% reduction in number of media queries.

I think it's worth it though, since we're likely to add more media queries in future as we add more components. Without this, our filesize might grow more substantially as we do so.

I also tweaked our Stylelint rules to guard against media queries that use explicit values rather than the `grav-breakpoint()` function (or a SASS variable). While we didn't have any instances yet, I think this is worthwhile to avoid any discrepancies creeping in in future. Again, keeping our media query code more consistent will help the PostCSS plugin better optimise our CSS code.

Finally, I wanted to add a note to our coding standards about media queries and... _well_, I got a bit carried away and tidied up and expanded the whole lot. :-)


packages/gravity-ui-web/docs/coding-standards.md
```diff
@@ -0,0 +1,299 @@
...
+
+* Use UNIX-style line endings (i.e. `\n`).
+* Indentation using 2 space characters.
+* Be UTF-8 encoded.
+* Have a blank line at the end.
+
+Note, we have an [`.editorconfig` file](https://editorconfig.org/) in the root of this repository. Most IDEs support this and will therefore apply the above rules automatically.
+
+## File names
+Source files should have lowercase, kebab-case filenames.
```
dw-buildit: Maybe we should use fs-lint to enforce this?

packages/gravity-ui-web/docs/coding-standards.md
```diff
@@ -0,0 +1,299 @@
...
+  @extend .grav-other-thing;
+
+  // SASS mixin @includes that do NOT use content blocks
+  @include grav-fancy-mixin(...);
+
+  // CSS custom properties
+  --foobar-barfoo: 42;
+
+  // SASS variable declartions
+  $foobar: 666;
```
dw-buildit: if we declare sass variables after custom properties how would we assign the sass variable to the custom property in the block?

don't know that there's a use case for it but seems like we could just put them the other way round and not have the problem?

james-nash: Good point! I'll update the rule and doc accordingly!

packages/gravity-ui-web/docs/coding-standards.md
james-nash: I didn't even know such a thing existed. Sounds like a good idea. Mind raising an issue (or even PR) to do it?

PR #336, opened 2019-07-29
Feature/padding utilities

danbull:

adds a suite of padding utility classes

Description
Currently we have a problem with the footer not having enough space above it. After much discussion in person and on this PR #323 I've added a function that will generate a suite of padding utility functions. I've added a class to the main element which will add the padding we need. They should be useful in the future for other cases where extra/less padding is required.

**Checklist**
### SASS

### DIFF
--- a/packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
+++ b/packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
@@ -1,6 +1,6 @@
 {%- include 'organisms-page-header' -%}
 {%- block pageContent %}
-<main>
+<main class="grav-u-pb-xl">
   {% block mainContent -%}
     {#
       Just some dummy example content.
--- a/packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
+++ b/packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
@@ -1,6 +1,6 @@
 {%- include 'organisms-page-header' -%}
 {%- block pageContent %}
-<main class="grav-u-pb-xl">
+<main>
   {% block mainContent -%}
     {#
       Just some dummy example content.
--- a/packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
+++ b/packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
@@ -1,0 +1,11 @@
+$grav-sides: (top, bottom, left, right);
+$grav-spaces-map: ('xs': $grav-sp-s,'s': $grav-sp-s, 'm': $grav-sp-m, 'l': $grav-sp-l, 'xl': $grav-sp-xl, 'xxl': $grav-sp-xl);
+
+// Generate utility classes for available spacing sizes
+@each $name, $space in $grav-spaces-map {
+  @each $side in $grav-sides {
+    .grav-u-p#{str-slice($side, 0, 1)}-#{$name} {
+      padding-#{$side}: #{$space} !important;
+    }
+  }
+}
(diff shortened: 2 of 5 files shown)


packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
```diff
@@ -0,0 +1,11 @@
+$grav-sides: (top, bottom, left, right);
+$grav-spaces-map: ('xs': $grav-sp-s,'s': $grav-sp-s, 'm': $grav-sp-m, 'l': $grav-sp-l, 'xl': $grav-sp-xl, 'xxl': $grav-sp-xl);
```
c1rrus: Should we have a zero option too for removing a padding?

Also, I wonder if it might make sense to move this map up into the settings layer. I can imagine it being useful elsewhere too - e.g. of we ever add margin util classes. You might also want to consider renaming it to `$grav-sp-map` to align woth our naming conventions.

packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
```diff
@@ -1,6 +1,6 @@
 {%- include 'organisms-page-header' -%}
 {%- block pageContent %}
-<main>
+<main class="grav-u-pb-xl">
```
c1rrus: Hmm.. I know we agreed to add padding utils, but were they the solution we settled on for space between `main` and page footer?

I thought removing a current exception to the `* + *` rule, so that it applies to these elements was still being considered too. Or not?

packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
```diff
@@ -0,0 +1,11 @@
...
+
+// Generate utility classes for available spacing sizes
+@each $name, $space in $grav-spaces-map {
+  @each $side in $grav-sides {
+    .grav-u-p#{str-slice($side, 0, 1)}-#{$name} {
+      padding-#{$side}: #{$space} !important;
```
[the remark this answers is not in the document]
danbull: I noticed we're not really using it on utility classes but it's usually convention to add it to your trumps. You want to guarantee that it will win or else the class shouldn't be used. Happy to remove it though.

packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
danbull: Yep a zero one would be useful, can add that.  Sure can move it to settings.

packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
danbull: My main motivation for adding the utility classes was to fix the footer problem. It seemed an ok solution to me and is what I've seen done on other sites. Can discuss when you're back ;)

packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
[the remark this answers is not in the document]
danbull: I did a bit of digging around and it does seem common practice to have `!important` on utilities - they should be used to enforce a single rule but you wouldn't want that rule to be overridden or else the class should be removed from the HTML. But if you think that's not in line with Gravity's standards then I can remove.

packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
danbull: Added a none value to the map

packages/gravity-ui-nunjucks/src/_patterns/04-templates/00-page-layouts/00-basic-page.njk
danbull: So, I decided to remove this from this PR for now and we can fix it in another PR. I still have a requirement for these classes now aside from this issue.

packages/gravity-ui-web/src/sass/06-utilities/_padding.scss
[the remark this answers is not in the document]
james-nash: No it's OK. Makes sense in this instance.

Did you run the sass linter? If it flags up the use of `!important` you may need to add a `// stylelint-disable-next-line ... ` comment to exclude this instance.

PR #339, opened 2019-08-23
Miscellaneous layout changes

### LINKED ISSUE #133: Flex on body causing undesirable side effects

### LINKED ISSUE #182: Section padding inconsistent
**Describe the bug**
When implementing page layout following the homepage template: http://style.buildit.digital/?p=templates-homepage there is no padding bottom on the last section above the footer.

In the demo in pattern lab no sections are rendered so the problem isn't visible. `<section class="grav-c-hero">` is fine.

On the website there is a `<section class="grav-o-container-banner grav-u-text-centered">` as the final section of the page which is also fine because it adds the padding correctly.

If however you just have a regular `<section>` as the last child of `<main>` it will not add padding and so will be tight against the footer:

screenshot 2019-02-04 at 16 19 29

### LINKED ISSUE #333: Change default layout proposal
Currently, Gravity removes all padding and margin from the `body` element (so everthing is "full bleed" by default). The intention is that consumers will use the `grav-o-container` class to box in any top-level elements that should not run up to the edge of the viewport, and also not extend beyond a sensible max width on very wide viewports.

A consequence of this is that a simple page with plain mark-up (e.g. something you might get by rendering markdown to an HTML file) won't be very usable until the author manually boxes in things.

This may seem minor, but it is somewhat at odds with the Gravity ethos applied in most other places which is to have sensible styling defaults that work with plain HTML. Ideally, for simple pages, no classes should be necessary at all. Some well structured HTML should yield a usable (though perhaps non-exciting) page layout.

To remedy this, I propose the following:

* We make the `body` element behave similar `.grav-o-container` by default. I.e. has a small margin on narrow viewports and a max-width that gets centered on wide viewports. We'd also need to some vertical margin above and below - probably `$grav-sp-m`.
    * This will mean that plain, class-less HTML won't have text that buts up against the edges of the viewport.
    * I mocked it up in a CodePen here: https://codepen.io/cirrus/pen/XvmEoW - note how `body` can be styled and sized just like any other element.
* We then have an object class that undoes this behaviour and thus makes it behave just like it does today. Perhaps call it `grav-o-full-bleed-body`?
    * This essentially makes the current layout behaviour opt-in. Consumers doing so are expected to know what they are doing and make use of `grav-o-container` as needed.
* We'll need to update the layout documentation accordingly.
* We should also add some extra rules to `debug.css` to help people avoid issues.
    * For example, we could flag the use of `grav-o-container` outside of `grav-o-full-bleed-body`.

This will of course be a **breaking change** since existing consumers will need to add the new class or risk their page layouts breaking. But, if we can get this in as part of Gravity NEXT, I think it's Ok since we already have a whole bunch of breaking changes so one more won't be too bad.

**Describe alternatives you've considered**
Not exactly an alternative, but rather a tweak of the above. Considering that `grav-o-container` will now only make sense within `grav-o-full-bleed-body` (or whatever name we settle on), perhaps we should change it to be a BEM-style _element_ class, e.g.: `grav-o-full-bleed-body__content`? That might help reinforce the relationship between the two and their intended usage.

TBH, I was never completely sold on the `grav-o-container` name anyway - it feels a bit too generic for what it is.

james-nash:
**Description**
Monster PR that implement a number of proposed and inter-dependent layout changes (and some other bits and bobs):

* **Full-bleed layout is now opt-in**. By default, `<body>` will have some padding and margin around it so that its contents don't touch the edges of the viewport and also have a sensible max-width on large viewports. This is, of course, a breaking change.
    * To restore the "old", full-bleed layout behaviour you can add the `.grav-o-full-bleed` class to `<body>`.
    * The old `.grav-o-container` class has been renamed to `grav-o-full-bleed__content` to better illustrate its relationship with `grav-o-full-bleed`.
    * Since we'd like page header and footer components to be flush with the top / bottom of the viewport, they have negative margins to counteract the `<body>` default padding. These get switched off when the `grav-o-full-bleed` class is used, so that header and footer will be correctly positioned in both scenarios. Similarly, their inner `<div>` autmatically adopt container behaviour when they appear within a full-bleed page.
    * Closes #333 
    * This change also removes the `body > *` exception to the lobotomised owl rule and thus solves the problem of a missing margin or padding above the page footer. Closes #182 
* **Sticky footer is configurable and more robust**. By default, the sticky footer effect still gets applied to `<body>`, however you now have the ability to disable that at build time via a SASS variable and also via a new `.grav-o-undo-sticky-footer` class for ad-hoc overrides.
    * To support SPA cases, where the page footer is not a direct descendent of `<body>` there is now also a `.grav-o-sticky-footer` class that can switch on this behaviour for arbitrary container elements.
    * The implementation of the sticky footer has been changed and should have less side-effects. It also no longer relies on the presence of a `<main>` element.
    * The implementation has been wrapped up a SASS mixin, so it can be easily re-used.
    * Closes #133 
* **Smart borders**. I made a few mixins for adding borders to containers. They have a "smart" option that makes them disappear if you apply a color scheme class to the same element. This can be useful for boxing in sections (e.g. a sidebar or similar). When that section doesn't have an explicit color scheme (and so, shares the same background color as its surroundings) it gets a visible border. But, if you apply a color scheme to it - which will change its background color - then the border disappears. The assumption is that the difference in background color is sufficient to visibly differentiate that container from its surroundings.
    * I was motivated to add this by the page footer. So far, unless you apply a color scheme to it (like we do on the Buildit website, so make it dark gray), there's no visible separation between it and the page content above. I've therefore given it a "smart" top border.
    * This is obviously a subjective design choice, so if you disagree, I'm happy to ditch it. However, even if we don't use it on the footer, I believe the smart border facility will be useful. I therefore suggest we keep the mixin in any case.
* **Removed `.grav-o-section` class**. All it did was apply a larger than usual margin-top. Probably should have killed it a long time ago. It's barely used and in light of all the other code & design changes redundant. If you really need a bespoke, bigger gap somewhere you can probably achieve it just as well with the recently added padding utility classes.
* **Adds margin utility classes.** This is useful for the homepage, where - as an exception - we want to remove the standard margin top that gets applied to `<main>` so that the hero is flush with the page header. Since we already have a full suite of padding util classes, I thought I may as well add all the equivalent margin ones too.
* **Standardises gap after page header**. The page intro component, used on story pages, previously had an extra large `margin-top` (which, strictly speaking, didn't follow Gravity's own layout conventions). On any pages where that component was not used, the gap above the page header's first sibling was the standard vertical gap, which itself varied, depending on that element's font size. When navigating between different pages this can cause the start of the page content to seamingly "jump" up and down. This change therefore removes the outer margin on the page-intro component (so that component now adheres to our usual conventions) and now sets an equivalent (and font-size indenpendent) top margin to any element that follows the page header.

To support testing the above changes, I have added a few page templates into the pattern library: "Simple HTML" (simulates the raw output you'd get from rendering a markdown file), "Header and footer" (simple HTML that has been topped and tailed by Gravity's page header and footer components), "Nested Root" (same as "Header and footer", but with everything wrapped in a top-level `<div>` so that they are not direct descendents of `<body>`.

I had previously been prototyping these in various CodePens, but it was getting a bit cumbersome to keep them all in sync, so doing it within our pattern library was easier. I've tried to keep each one as a seprate commit, so if we decide to ditch any of these ideas, we _should_ be able to rebase and cherry-pick just the ones we want to keep.


packages/gravity-ui-web/src/sass/04-objects/_sticky-footer.scss
```diff
@@ -0,0 +1,17 @@
...
+  // need to adjust our min-height accordingly:
+  .grav-o-full-bleed & {
+    min-height: 100vh;
+  }
+}
+
+.grav-o-undo-sticky-footer {
+  display: block;
+  min-height: none;
+}
```
danbull: Although it's not just just one property that's being overriden I'd say this may be better suited as a utility class - perhaps using a basic mixin (could be scaled up for other uses in the future). It does really lend itself more to a modifier (since it does one job and also only has one use case). A modifier with no block selector though is a bit awkward maybe? 🤔

dw-buildit:
@james-nash When i'm looking at this branch locally on pattern lab, i'm not getting any padding on the body of Templates and Pages. Do I need to run something? Or is this a bug?

james-nash:
@dw-buildit Hmm... it should _just work_TM

Maybe the rebase broke something. Let me try and, once it's all clean, I'll push it.

packages/gravity-ui-web/src/sass/04-objects/_sticky-footer.scss
james-nash: @dw-buildit Just had a chat about it and decided to go with a modifier. While it is an "odd" one in that you'd probably never use it together with the corresponding "block" class, it does feel like the most logical home for this. Hopefully, its also where developers might think to look for a way to undo the sticky footer behaviour.

james-nash:
Ok, just pushed the renamed "undo sticky footer" class: `..grav-o-sticky-footer--unstick`.

Issue #167, opened 2018-11-22
Expose version and build info in pattern library


james-nash:
**Is your feature request related to a problem? Please describe.**
When visiting our pattern library website http://style.buildit.digital/, there's no way of knowing which version of our library it corresponds to. You therefore have to poke around here on Github to look that up.

**Describe the solution you'd like**
The Pattern Lab site should expose the library version and build meta-data somehow.

We had a similar issue on the Buildit website and therefore added a feature to include such build meta-data in the output - see buildit/buildit#165. We could perhaps port the same thing over to this project.

Additionally, we may want to display the version number visually in Pattern Lab somehow.

james-nash:
Perhaps we can extract the build info stuff we did for the website into its own NPM package. Both the website and Gravity could then make use of that - and potentially other future projects as well.

james-nash:
This has been done in `next`.
<img width="731" alt="image" src="https://user-images.githubusercontent.com/16718916/71669417-7e781580-2d64-11ea-95cb-ddbe0d88605a.png">

PR #350, opened 2019-09-09
Miscellaneous documentation improvements

### LINKED ISSUE #106: Improve global look and feel of the pattern library
As an effort to make this pattern library easy to browse and consume, I'm opening this ticket as I think we can proceed with a series of small improvements that can largely affect the way we use, consume, and improve this tool. 

I'm trying to keep the conversation in here so we are aware of the decisions and what is needed to be done.

Here's a random list of things that I'd love to address:

1. **Particles** section should be revisited with a more precise intent and it should be documented, we are effectively expanding Atomic Design to include something that is not part of the definition of the design methodology and should be noted down. This seems to fall very much into the Design System area and the "let's put it here until we have a better place".
    1. with this in mind perhaps it's just a question of making this a more textual explanatory area and have a small introduction. Nothing wrong in having a simple md file that deals with it. (I have no idea how to achieve this, will need investigation)
    2. we are currently missing a show how responsive typography works via gravy/typi: we ought to showcase how the various sizes are working across various breakpoints and how they relate to each other size-wise. At the moment it is a bit tricky to understand and use them correctly. This is a very important bit IMO.
    3. widths/heights, variables, vertical rhythms... all part of the above problem, probably it's just a question of finding a way to display this info in a concise and simple way.
2. **Atoms**: 
    1. classless atoms: the list is annoyingly long and possibly useless for various reasons. Perhaps grouping all atoms in a more cohesive way could be the best way to approach this, an example could be how many other companies are doing this in pattern libraries by writing a excerpt from a book. Clearly the intention is the important bit, this type of markup is stuff that we would potentially use in parsed markdown data, so I think it's fair to be represented as such. As far as I can see, majority of these atoms might not be even be used as patterns via template inclusion as they require custom variables to be passed into them. 
    2. atoms with classes: although, as in the previous case, all these classes might not be used as patterns via template inclusion, their aggregation might be a bit more tricky. Specifically when it comes to buttons or special headings or lists.
3. **Templates**: We need to be more careful on the way templates are shown and their actual intention. We have an homepage that doesn't show anything between the header and the footer... not ideal.
4. **Pages**: this is valid for templates as well: these should be removed from the global listing, so I'd rather avoid having the "generic page" example at all (that is, avoid grouping via subfolders).

I'm completely happy to split this ticket in sub tickets as soon as we come to the conclusion that any of what we will discuss is worth investing time into.

### LINKED ISSUE #119: Document Gravity goals and principles in Pattern Lab
References #118

We have an initial set of goals and principles for Gravity. For now, they should be housed within Pattern Lab, as that's the easiest way to make them accessible to a wide audience.

Question to consider: Should this live under "Particles", or should we introduce a new top-level section. "Docs" perhaps?

james-nash:
**Description**
* Adds guide to Fractal for devs used to Pattern Library (addresses outstanding feedback from the recent migration to Fractal)
* Adds a documentation section to the pattern library
    * Updates layout and color system docs to match recent updates and moves them into the pattern library (so that they are easier to discover)
    * Adds Gravity's goals and principles to the pattern library (closes #119)
    * Displays version and environment name in pattern library (part of #167)
* Smartens up pattern library (closes #106)
    * Some light customisation to Fractal's default theme: Header uses our teal accent color
    * Overview page has a Gravity logo on it :-)
    * Pattern library now has a `robots.txt` to allow the production version and block staging, next and local dev ones from being indexed. (Similar to what we do for the Buildit website envs)


james-nash:
Thanks @dw-buildit . I've implemented your suggestions. Will merge this PR as soon as they automated checks pass.

PR #356, opened 2019-09-25
refactor: updates debug.css

james-nash:

Updates debug.css to bring it inline with recent changes to Gravity's CSS classes. This also adds
some errors for v2 classes that have been removed in v3.

I've also thrown in a mixin to display warning messages for deprecated classes. It's not being used yet, but could be useful in future if we ever intend to deprecate things a few releases before the actually get removed.

### DIFF
--- a/packages/gravity-ui-web/src/sass/debug.scss
+++ b/packages/gravity-ui-web/src/sass/debug.scss
@@ -91,12 +91,11 @@
 
   @if $version-to-be-removed {
     // Append message with version to be removed
-    $message: '#{$message} and is due to be removed in v#{$version-to-be-removed}.';
+    $message: '#{$message} and is due to be removed in v#{$version-to-be-removed}';
   }
-  @else {
-    // End the sentence
-    $message: '#{$message}.';
-  }
+
+  // End the sentence
+  $message: '#{$message}.';
 
   @if $migration-message {
     // Append migration message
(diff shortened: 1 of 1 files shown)


packages/gravity-ui-web/src/sass/debug.scss
```diff
@@ -80,21 +80,69 @@
...
 /* stylelint-enable scss/at-mixin-pattern */
 
 // ====================================================================
 // Tests for misuse of Gravity's styles, classes and components
 // ====================================================================
 
-.grav-o-container .grav-o-container {
-  @include highlight-error('Nesting of .grav-o-container is not allowed.');
+.grav-o-full-bleed__content .grav-o-full-bleed__content {
+  @include highlight-error('Nesting of .ggrav-o-full-bleed__content is not allowed.');
```
dw-buildit: typo `ggrav`

packages/gravity-ui-web/src/sass/debug.scss
```diff
@@ -80,21 +80,69 @@
...
+  $message: '#{$class} has been deprecated';
+
+  @if $version-to-be-removed {
+    // Append message with version to be removed
+    $message: '#{$message} and is due to be removed in v#{$version-to-be-removed}.';
+  }
+  @else {
+    // End the sentence
+    $message: '#{$message}.';
+  }
```
dw-buildit: You can save some chars:

```
  @if $version-to-be-removed {
    // Append message with version to be removed
    $message: '#{$message} and is due to be removed in v#{$version-to-be-removed}';
  }

  // End the sentence
  $message: '#{$message}.';
```

dw-buildit:
@james-nash Approved, although there are a couple of minor alterations you could make (typo etc)

PR #364, opened 2019-11-12
Automated a11y testing

### LINKED ISSUE #361: Add automated a11y testing
We should add some a11y testing to our UI components (after all, _Inclusive_ is one of Gravity's principles!).

Such tests could flag common pitfalls like:
* missing `alt` attributes on images
* missing associations beween `<label>` elements and their associated inputs
* empty links and buttons
* incorrect use of `aria-*` attributes
* insufficient color contrast
* etc.

This should help us ensure that our UI components individually are designed and built correctly and won't prevent downstream consumers from assembling accessible UIs from them.

1. Evaluate available a11y testing tools and pick an appropriate one.
    * [aXe](https://www.deque.com/axe/) and [Pa11y](https://pa11y.org/) are obvious candidates to consider, but there may be more.
1. Set it up, so that every UI component in our pattern library gets tested (and ones we add in future automatically get tested too).
    * Tools like aXe and Pa11y tend to operate on HTML and/or hosted pages (which they then access via a headless browsers), so we can probably set this up within `packages./gravity-ui-nunjucks/` and have the tools access the per-component pages that Fractal already generates for us. Note though that we want to specifically test _our_ component HTML, _not_ Fractal's pattern library UI. The "preview" pages it creates (e.g. http://localhost:3000/components/preview/superseded ) look appropriate. So, if we could get our chosen tool to iterate over all the preview pages, that would be ideal.
1. Add an NPM script to the `package.json` to make it easy to run the tests. Perhaps `test:a11y`, so we can invoke them by doing `npm run test:a11y`.
    * It probably makes sense to also add a `test` script, which for now would just run the a11y tests. In time, if we add other kinds of tests, they can be appended. That way `npm run test` will run _all_ tests, where as `npm run test:a11y` will only run the a11y tests
    * I _think_ doing this will also ensure that our Travis CI will execute these tests. It already run's the root package's `test` script, which calls `lerna run test` and that in turn will run the `test` script for all sub-packages that have one. Maybe verify this is indeed working as expected and, if not, do whatever's needed to fix it. The CI _should_ be running the a11y tests.
1. Update docs as needed
    * Update the `README.md` with instructions on how to run the tests.
    * If you think it's worthwhile to provide additional details or intructions, then add to or update the docs as needed

At this point, I'd say the work is done.

If the tests highlight lots of errors (and thus cause CI builds to fail), we may need to temporarily disable the a11y tests on CI builds and raise new Github issues to get the a11y bugs fixed. On the other hand, if there's only a few bugs and they're trivial to fix, then it might be easier to just fix them as part of this work.

Elohina:
Closes #361

**Description**
Add automated accessibility tests to all the components of the UI library through the preview URL's of each one. This was possible using axe-webdriverjs, the axe-core API for Selenium's WebDriverJS. 

**Checklist**

1. Evaluate available a11y testing tools and pick an appropriate one.
- [x] It was selected axe-core since it was easy to find documentation and examples. Also, axe-webdriverjs let us run integration test to every component and complete pages.
2. Set it up, so that every UI component in our pattern library gets tested (and ones we add in future automatically get tested too).
- [x] Every component is analysed through the preview URL
- [x] The test suite iterates over an array of components generated from the component's directory.
- [x] Every component test is autogenerated, so new components placed in `gravity-ui-nunjucks/src/components` will be tested.
3. Add an NPM script to the package.json to make it easy to run the tests.
- [x] `npm run test:a11y` command was added to the `package.json` in order to run the a11y test.
4. Update docs as needed.
- [x] README.md was updated with the instructions to run a11y tests.


packages/gravity-ui-nunjucks/__tests__/a11y.spec.js
```diff
@@ -0,0 +1,73 @@
...
+  'designit',
+  'icon',
+];
+
+describe('A11y tests', () => {
+  let driver;
+  const components = getComponentsNames(pkgPaths.srcComponentsPath(), excludedFiles, '.njk');
+
+  components.forEach((component) => {
+    fdescribe(`${component} component`, () => {
```
FiF0o: I think we need to replace fdescribe to describe as it will lock the tests to this describe block only.

james-nash:
@Elohina Great work! I've finally gotten around to checking out the branch and trying it out.

I will take a closer look at the code in a moment, but wanted to share a few initial observations:

* Installing ChromeDriver was a bit fiddly - on my Mac (running macOS Catalina) I had to grant a few permissions before it would run because it's not by a "verified developer". Also, it seems to be closely tied to the version of Chrome you have installed. I had a mismatch the first time and it refused to run. I had (very quick) look around and this NPM package might help the setup a little easier: https://www.npmjs.com/package/chromedriver
* Am I right in assuming the ChromeDriver requires a compatible version of Chrome to already be installed? While this may be fine for most developers, I wonder whether that would be difficult later on when we want to run on TravisCI. This might not be something we need to resolve right now in this PR (as discussed, let's get this merged and focus on fixing existing a11y issues before we start running these tests on CI), but might be something we need to investigate later on. I wonder if we can use some kind of headless browser (like [Puppeteer](https://www.npmjs.com/package/puppeteer)) instead of Selenium + Chrome. The ideal experience would be that `npm install` downloads and sets up _everything_ that's needed, so that CI works fine and developers don;t need to manually install things on their machines.
* The pattern library needs to be running locally (ideally, via `npm start` in the monorepo's root so that the CSS is also compiled) _before_ you run the a11y tests. It would be good to mention that in the instructions in the `README`.
* Related to the above, when you don't have the pattern library running but run the tests, they fail (as expected) but not with a helpful error. Currently they report:

```
Ensures each HTML document contains a non-empty <title> element
 Documents must have <title> element to aid in navigation
 Help: https://dequeuniversity.com/rules/axe/3.4/document-title?application=webdriverjs

 Ensures every HTML document has a lang attribute
 <html> element must have a lang attribute
 Help: https://dequeuniversity.com/rules/axe/3.4/html-has-lang?application=webdriverjs
```
...which I think creates the false impression that it's managing to access the pattern library but finds errors in the HTML. Would it therefore be possible to identify when the pattern library is unavailable and report that (e.g. `Cannot connect to http://localhost:3000/. Did you forget to run the pattern library locally?`? _Or_, and perhaps this would be better, update the gulp test task to launch the pattern library and the tests in parallel (but - you might need to watch out for race conditions where the tests try to access the pattern library before it's finished starting up).

packages/gravity-ui-nunjucks/README.md
```diff
@@ -52,6 +52,16 @@ npm start
...
+#### Run accessibility tests
+Runs component's accessibility tests using [axe-webdriverjs](https://github.com/dequelabs/axe-webdriverjs).
+
+First download and install the lastest `chromedriver` from [here](https://chromedriver.storage.googleapis.com/index.html).
+
+[See more on Webdriver requirements and setup](https://selenium.dev/documentation/en/webdriver/driver_requirements/).
+```
+npm run test:a11y
+```
+
```
james-nash: As per my [other comment](https://github.com/buildit/gravity-ui-web/pull/364#issuecomment-555516579), it would be good to add a note here to explain that the pattern library needs to be running before you can launch these tests.

Unless you update the Gulp task to do so automatically... ;-)

packages/gravity-ui-nunjucks/__tests__/a11y.spec.js
```diff
@@ -0,0 +1,73 @@
...
+];
+
+describe('A11y tests', () => {
+  let driver;
+  const components = getComponentsNames(pkgPaths.srcComponentsPath(), excludedFiles, '.njk');
+
+  components.forEach((component) => {
+    describe(`${component} component`, () => {
+      beforeEach((done) => {
+        driver = new selenium.Builder()
```
james-nash: Is it possible to move the setup steps into a `beforeAll()`, so that the browser is only launched once? Presumably the setup for each test only needs to be constructing the URL and calling `driver.get(...)` to navigate to the new page.

This might speed up the test runs a bit.

packages/gravity-ui-nunjucks/__tests__/_helpers.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+   * Takes a set of objects and returns an array of formatted error messages.
+   *
+   * @param  {array} violations Set of violation objects.
+   *
+   * @return {array} Array of formatted messages.
+   */
+const getViolations = (violations) => violations
+  .map((violation) => `${violation.description}\n ${violation.help}\n Help: ${violation.helpUrl}\n`);
```
james-nash: Maybe update this to also include details about which element is causing the violation.

packages/gravity-ui-nunjucks/gulpfile.js
```diff
@@ -66,8 +68,14 @@ function clean() {
 }
 clean.description = 'Deletes all intermediated and final build output files.';
 
+const test = gulp.series(
+  patternLibraryTasks.startServer,
```
james-nash: 🎉

packages/gravity-ui-nunjucks/__tests__/a11y.spec.js
Elohina: `beforeAll()` was added and changed from scope in order to launch the browser once. Also it was changed the library. Instead of using Selenium, now we are using Puppeteer. So, installing a Chrome driver is not required.

packages/gravity-ui-nunjucks/__tests__/_helpers.js
Elohina: Elements causing a11y violations are now included in the tests results 😄

PR #394, opened 2020-03-23
The Gravity 3 Motherlode

james-nash:
After more than a year of work, I think it's time to merge the Gravity 3 work we've been doing in `next` up into `develop`, so that we can then release it. 🎉 

Looking through the remaining issues, there's some pattern library tidy-ups we want to do, but I don't think they'd be likely to introduce breaking changes to the SASS/CSS code. So, no reason to hold back the release of `v3.0.0` just for those.

There shouldn't be any need to review the changes in this PR as each one will have been individually done as a PR on `next` and thus reviewed at the time. But, reviewers are of course welcome to do whatever (spot) checks they want.


danbull (approved):
🍾

PR #600, opened 2020-08-06
Helvetica and spacing

allancorbett:
**Description**
Implements Helvetica and adjusts spacing between objects.


packages/gravity-ui-web/src/sass/05-components/03-organisms/_hero.scss
```diff
@@ -36,6 +36,7 @@
     }
 
     h1 {
+      max-width: $grav-page-content-max-width - ($container-horizontal-max-margin * 2);
```
james-nash, line 39: Why is this needed?

Shouldn't headings just have their default `display: block;` behaviour and fill out whatever horizontal space they're given by their parent?

If a certain consumer of Gravity, like the website, need to tweak things they should do so in their code. Otherwise, we might risk breaking things in other consumers - e.g. Cycle 7 website.

packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
```diff
@@ -7,7 +7,10 @@ $grav-page-heading-trigger-breakpoint: grav-breakpoint(medium);
 // width ($grav-page-content-max-width) themselves, where necessary.
 body > header,
 .grav-c-page-header {
+  position: sticky;
```
james-nash, line 10: Hmm... I'm not sure about doing this here. That would mean any project using Gravity will automatically get a sticky header (and it will be sticky across all screensizes too). Is that really what we want?

If this is intended for the Buildit website, then I'd suggest putting those rules in the website's own CSS. It's fine for downstream projects like the website to tweaks some of Gravity's defaults to suit their specific needs, but I'd be weary of doing something like this in Gravity itself.

That being said, this is not a hard "no". If you feel strongly this is how all Gravity page header should be, then this is the right place to do it.

james-nash: Also, whether its here or in the website's code, don't forget to give the header a solid background colour. Right now, the page content shines through and it becomes unreadable...
image

README.md
```diff
@@ -1,45 +1,53 @@
...
 ## Development
 
 ### Initial setup
+
 After an initial clone, or if any of the packages' dependencies have changed you should run the following **in the repo's root directory**:
 
 ```
 npm install
-npm lerna:bootstrap
+lerna bootstrap
```
james-nash, line 25: Actually, the instructions were correct as they were.

`lerna bootstrap` would only work if you happen to install Lerna globally, which I'd be reluctant to force people to do (I find global NPM things just clutter up your computer and sometimes lead to weird version conflicts when they differ to a locally installed copy in a project)

README.md
```diff
@@ -1,45 +1,53 @@
...
 ```
 
 This will install dependencies for all packages in the repo and configure any local cross-references between them.
 
+You may need to install [Lerna](https://lerna.js.org/) globally using:
+
+```
+npm lerna:bootstrap
```
james-nash, line 33: If you revert the bit above, you can get rid of this too.

FWIW, this would just run `lerna bootstrap` - but using the copy of lerna installed into Gravity's `node_modules`. To install lerna globally, you'd need to do `npm install -g lerna`.

packages/gravity-ui-web/src/sass/05-components/03-organisms/_hero.scss
allancorbett: Correct, removed.

packages/gravity-ui-web/src/sass/05-components/03-organisms/00-page-structure/_00-page-header.scss
allancorbett: Correct!

README.md
allancorbett: 👍

